PR #1, opened 2024-02-09
API specification

jan-zabloudil:
**General notes & TBD** 

1.  Consider whether API versioning is necessary. Since this tool is only for internal use, it might not be crucial at this stage.

2. Nested resources are not used because of simple application structure.

✅ /projects/{project_id}/apps
✅ /apps/{app_id}
❌ /projects/{project_id}/apps/{app_id}

3. User sign-up and sign-in functionality do not have dedicated endpoints as Supabase Google SSO is used. For sing-up/login action client has to redirect users to `https://<project>.supabase.co/auth/v1/authorize?provider=google`.

4. To create an admin user, first, the user needs to sign up. Then, the admin flag can be added to the user via the Supabase SQL editor by executing the following command: `select set_claim('user_id', 'is_admin', 'true')`. ❗If the 'is_admin' role is removed for a user, all user's tokens have to be invalidated. More information on [custom claims](https://github.com/supabase-community/supabase-custom-claims?tab=readme-ov-file).

**Endpoint notes:**

`/organization/settings`

GitHub token can be set at organization level. For the application level, users can choose to either provide only the repository URL, in which case the organization token will be used, or they can provide a token specifically for that repository.

`/invitations/{invitation-hash}/confirm`

- When a user is invited to a project, they need to click on the link sent in the email and confirm the invitation. If the user doesn't have an account yet, the client should redirect them to the sign-up page. After signing up, the user will be added to the project.
- The endpoint will provide information in the response about whether the user is already signed up or not.
- Another option would be to implement a GET endpoint and the link would be sent directly in the email. In this case, ensure that the token is [not leaked in the referrer header](https://shahjerry33.medium.com/password-reset-token-leak-via-referrer-2e622500c2c1).

`/apps/{app-id}/releases/actions`

- When a release is created, users have the option to specify whether a GitHub release should also be created and if a Slack message should be sent. 
- Alternatively, users can trigger GitHub release creation or send a Slack message later by calling the `apps/{app-id}/releases/actions` endpoint. It is necessary to discuss whether this endpoint adheres to RESTful principles or actions should be invoked by calling `/releases/{release_id}` patch endpoint.
- Certain constraints can be applied to these actions; for instance, creating a GitHub release may only be allowed if the release has been deployed to production.

`/projects/{project-id}/apps`

The endpoint also enables to define an URL of each deploy environment (dev, staging, prod). When creating a release, users have the option to specify the environments to which the release has been deployed. With this information, a link to the deployed app can be included in the Slack message.

`/apps/{app-id}/releases`

- When creating or updating a release, users have the option to provide a link to a GitHub release. If the creation of a release triggers the creation of a GitHub release, the GitHub release link is added automatically.
- Currently, the REST API does not offer functionality for triggering actual deployment to the development, staging, or production environments. Because of that, information about the environments (where the release is deployed) is simply stored in a single array field. This functionality can be expanded in the future when actual deployment triggering is implemented.
- Source code can be linked by providing a git tag. For further application improvements, the source code could also be linked by providing a specific commit, or it could be possible to create a new git tag directly from the app.

`/projects/{project-id}/templates`

Since each project can use only one message template at the time, it's worth considering whether it makes sense to create a separate message templates entity. Alternatively, each project could simply have a message template field.

`/tokens/refresh`

Should the refresh token be sent in the authorization header or in the request body?


bolshakoff:
If Mara approves it, I don't mind merging it; want to see the whole app in action ASAP, and then iterate via Issues, etc...

api-doc.yaml
```diff
@@ -0,0 +1,1635 @@
...
+          type: string
+          example: 'This project is about...'
+        notifications:
+          type: object
+          properties:
+            slack_channels_ids:
+              type: array
+              items:
+                type: string
+                example: [ "C065E66TZ36", "C065E66TZ38" ]
```
jan-zabloudil, line 1154: `slack_channels` -> `slack_channels_ids`

Channel name is not good identificator, therefore user should provide slack channel(s) id.

api-doc.yaml
```diff
@@ -0,0 +1,1633 @@
...
+          type: string
+          example: 'Awesome project'
+        description:
+          type: string
+          example: 'This project is about...'
+        notifications:
+          type: object
+          properties:
+            slack_channel_id:
+              type: string
```
jan-zabloudil, line 1151: Originally, users were able to submit an array of Slack channel names. However, I made the following changes:
* Slack channel ID instead of the name, as it serves as a better identifier.
* Only one Slack channel can be provided. Since STRV typically has one channel per project, it is sufficient for the MVP and will streamline implementation.

PR #4, opened 2024-02-25
[DSK-27] Users - read & delete endpoints

jan-zabloudil:
~~Supabase go client was also updated to provide new methods:~~
~~https://github.com/jan-zabloudil/supabase-go/pull/2~~

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -1137,9 +1137,9 @@ components:
           example: 'Jan Zabloudil'
         avatar_image_url:
           type: string
-        is_admin:
-          type: boolean
-          example: false
+        role:
+          type: string
+          example: admin
         created_at:
           type: string
           format: date-time
--- a/repository/model/user.go
+++ b/repository/model/user.go
@@ -6,6 +6,7 @@ import (
 	svcmodel "release-manager/service/model"
 
 	"github.com/google/uuid"
+	"github.com/nedpals/supabase-go"
 )
 
 func ToSvcUser(id, email string, roleString, name, picture any, createdAt, updatedAt time.Time) (svcmodel.User, error) {
@@ -35,4 +36,27 @@ func ToSvcUser(id, email string, roleString, name, picture any, createdAt, updat
 
 	u.Email, u.CreatedAt, u.UpdatedAt = email, createdAt, updatedAt
 	return u, nil
+}
+
+func ToSvcUsers(dbUsers []supabase.AdminUser) ([]svcmodel.User, error) {
+	users := make([]svcmodel.User, 0, len(dbUsers))
+
+	for _, dbUser := range dbUsers {
+		user, err := ToSvcUser(
+			dbUser.ID,
+			dbUser.Email,
+			dbUser.AppMetaData["role"],
+			dbUser.UserMetaData["name"],
+			dbUser.UserMetaData["picture"],
+			dbUser.CreatedAt,
+			dbUser.UpdatedAt,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		users = append(users, user)
+	}
+
+	return users, nil
 }
--- a/service/model/user.go
+++ b/service/model/user.go
@@ -7,6 +7,13 @@ import (
 	"github.com/google/uuid"
 )
 
+type UserRepository interface {
+	ReadForToken(ctx context.Context, token string) (User, error)
+	Read(ctx context.Context, id uuid.UUID) (User, error)
+	ReadAll(ctx context.Context) ([]User, error)
+	Delete(ctx context.Context, id uuid.UUID) error
+}
+
 var AnonUser = &User{}
 
 type User struct {
@@ -17,10 +24,6 @@ type User struct {
 	Role      UserRole
 	CreatedAt time.Time
 	UpdatedAt time.Time
-}
-
-type UserRepository interface {
-	ReadForToken(ctx context.Context, token string) (User, error)
 }
 
 func (s *User) IsAdmin() bool {
--- a/repository/user.go
+++ b/repository/user.go
@@ -7,6 +7,7 @@ import (
 	"release-manager/repository/utils"
 	svcmodel "release-manager/service/model"
 
+	"github.com/google/uuid"
 	"github.com/nedpals/supabase-go"
 )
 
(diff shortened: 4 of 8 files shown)


repository/utils/utils.go
```diff
@@ -0,0 +1,59 @@
...
+			return fmt.Errorf("%w: %s", reperr.ErrUserAuthenticationFailed, err.Error())
+		case 404:
+			return fmt.Errorf("%w: %s", reperr.ErrUserAuthenticationFailed, err.Error())
+		}
+	}
+
+	return fmt.Errorf("%w: %s", reperr.GeneralErr, err.Error())
+}
+
+func WrapSupabaseAdminErr(err error) error {
```
CermakM, line 47: I don't understand the name of this function. Why is this `AdminErr`? I get DBErr and AuthErr from the previous functions, but here it's not immediately apparent to me.

repository/model/users.go
CermakM, line 1: Singular `user.go`.

supabase/config.toml
CermakM, line 1: Singular `user.go`.

CermakM (changes requested):

repository/utils/utils.go
jan-zabloudil: Auth endpoint is called in order to authenticate user. User’s bearer token is provided in the authorization header.
**Error responses are following:** 

- 401 = invalid bearer token -> ReleaseManager returns 401
- 404 = user not found (valid token but user no longer exists in the database) -> ReleaseManager returns 401.

---
Admin endpoins are called in order to retrive users. Supabase secret key must be provided to access auth data.
**Error responses are following:**
- 401 = cannot access resource (invalid supabase secret key provided or only supabase anon key provided) -> ReleaseManager returns 500 (cannot access db because bad credentials were provided in the config)
- 404 = user not found (no user found for provided uuid) -> ReleaseManager returns 404

I needed to distinquish between those cases to return correct response. 
And I find better to create separate function for Auth and Admin than checking error text as well. I find it better to rely only on returned status code.

jan-zabloudil:
All review comments were addressed, lot of changes were made because i switched to `go.strv.io/net`.

This branch is rebased onto #12. So once this PR is approved, I am going to merge both branches, there is no point of doing review for #12 separately.

Sorry that commits are squashed already.

api-doc.yaml
```diff
@@ -60,7 +60,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/OrganizationSettings'
         '400':
-          $ref: '#/components/responses/ForbiddenErrorResponse'
+          $ref: '#/components/responses/BadRequestErrorResponse'
         '401':
```
jan-zabloudil, line 64: Not related to this PR. Just fix of error I noticed in openapi.

repository/model/user.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"time"
+
+	svcmodel "release-manager/service/model"
+
+	"github.com/google/uuid"
+	"github.com/nedpals/supabase-go"
+)
+
+func ToSvcUser(
```
CermakM, line 12: Use 1 line per argument or all args in a line.

repository/model/user.go
```diff
@@ -0,0 +1,61 @@
...
+	id, email string,
+	isAdmin, name, picture interface{},
```
CermakM, line 14:
```suggestion
	isAdmin, name, picture any,
```

transport/middleware.go
```diff
@@ -1,19 +1,80 @@
...
 
-func (h *Handler) recoverPanic(next http.Handler) http.Handler {
+func (h *Handler) auth(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		defer func() {
-			if err := recover(); err != nil {
-				w.Header().Set("Connection", "close")
-				WriteServerErrorResponse(w, r, fmt.Errorf("%s", err))
+
+		w.Header().Add("Vary", "Authorization")
```
CermakM, line 18: I believe you can get rid of Vary, browsers don't really support it and cache control happens based on Cache-Control or status response.

service/model/user.go
```diff
@@ -0,0 +1,25 @@
...
+	ReadAll(ctx context.Context) ([]User, error)
+	Delete(ctx context.Context, id uuid.UUID) error
+}
+
+type User struct {
+	ID        uuid.UUID
+	Email     string
+	Name      string
+	AvatarUrl string
+	IsAdmin   bool
```
CermakM, line 22: I'd prefer `UserRole` custom type here, it's more flexible, but you can do that in a separate PR.

cmd/api/main.go
```diff
@@ -2,85 +2,51 @@ package main
...
+	timex "go.strv.io/time"
+)
 
 func main() {
-	cfg := loadConfig()
+	ctx := context.Background()
+	cfg := loadConfig(ctx)
 
-	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: cfg.LogLevel}))
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: cfg.LogLevel}))
```
CermakM, line 24: Logs are usually logged to Stderr

repository/errors/errors.go
```diff
@@ -0,0 +1,12 @@
+package errors
+
+import "errors"
+
+var (
+	UnknownErr                                        = errors.New("supabase: unknown error")
+	ErrEntityNotExists                                = errors.New("supabase: entity does not exist")
+	ErrUserAuthenticationFailed                       = errors.New("supabase: user authentication by bearer token unsuccessful")
+	ErrAccessToResourceDenied                         = errors.New("supabsse: access to resource denied")
```
CermakM, line 9:
```suggestion
	ErrAccessToResourceDenied                         = errors.New("supabase: access to resource denied")
```

transport/context.go
CermakM, line 1: I would put this file until transport/util

CermakM (changes requested):

service/model/user.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+var AnonUser = &User{}
+
+type User struct {
+	ID        uuid.UUID
+	Email     string
+	Name      string
+	AvatarUrl string
+	Role      UserRole
```
jan-zabloudil, line 24: UserRole type added (see /services/model/user_role).

Also all user methods such as isAdmin, isAnon were moved into service layer.

service/model/user_test.go
jan-zabloudil, line 1: Tests added for user and user_role

transport/middleware.go
```diff
@@ -1,19 +1,79 @@
...
+				WriteInvalidAuthenticationResponse(w, err)
+				return
+			default:
+				WriteServerErrorResponse(w, err)
+				return
 			}
-		}()
+		}
+
+		r = utils.ContextSetUser(r, &user)
```
jan-zabloudil, line 47: The service User object is now stored in the context.
I believe that storing the User struct in the context is acceptable, the object is not too big.

CermakM (approved):
Good job! Let's merge this. 💪

PR #11, opened 2024-03-03
[DSK-32 ] Release message templates - CRUD endpoints

jan-zabloudil:


jan-zabloudil:
The endpoint URL is `/templates, user provides the template type and then the template data. If new template type is introduced in the future, there will be no breaking changes.

```
{
    "type":"release_message",
    "template_data": {
        "title":"New Release",
        "text":"It's here!",
        "includes":["changelog","deployments"]
    }
}
```

In database, template data are saved in one JSON property. Why?
- No migration needed if new template type is introduced
- Each project will have its own template and it is handy that the whole template can be saved to one db column (projects.template)

Snímek obrazovky 2024-03-17 v 22 08 30

Currently, the codebase is not prepared to handle more than one template type. However, I believe it is better to implement a simpler solution now and then refactor it once there is a real need for a new template type.

transport/utils/custom_validations.go
```diff
@@ -0,0 +1,20 @@
...
+	"fmt"
+
+	svcmodel "release-manager/service/model"
+
+	"github.com/go-playground/validator/v10"
+)
+
+func ValidateReleaseMessageType(fl validator.FieldLevel) bool {
+	return fl.Field().String() == svcmodel.ReleaseMessageTemplateType
+}
```
jan-zabloudil, line 13: Custom validation is added. I could add also do 

```
Type  *string `json:"type" validate:"required,eq=release_message"`
```

In this case, I couldn't pass the variable from the model and had to hard-code the value instead.

jan-zabloudil:
I've realized that a slightly different approach is needed. I am going to refactor the code.

jan-zabloudil:
The endpoint URL is `/templates`, user provides the template type and then the template data. 
```
{
    "type":"release_message",
    "template_data": {
        "title":"New Release",
        "text":"It's here!",
        "includes":["changelog","deployments"]
    }
}
```

In database, template data are saved in one JSON field. Why?
- No migration needed if new template type is introduced
- Each project will have its own template and it is handy that the whole template can be saved to one db column (projects.template)

![Snímek obrazovky 2024-03-17 v 22 08 30](https://github.com/j

service/model/template.go
```diff
@@ -0,0 +1,48 @@
...
+	ReadAll(ctx context.Context) ([]Template, error)
+	Read(ctx context.Context, id uuid.UUID) (Template, error)
+	Update(ctx context.Context, t Template) (Template, error)
+	Delete(ctx context.Context, id uuid.UUID) error
+}
+
+type Template struct {
+	ID         uuid.UUID
+	Type       TemplateType
+	ReleaseMsg ReleaseMessage
```
jan-zabloudil, line 26: The Template struct is currently directly linked to the concrete type ReleaseMessage. My initial goal was to abstract this relationship to simplify the process of adding new template types in the future. 

But it made the code more complex. In the end, I chose to keep things simple. It can be refactored if there will be need for new template type in the future.

repository/model/template.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	svcmodel "release-manager/service/model"
+
+	"github.com/google/uuid"
+)
+
+type Template struct {
+	ID         uuid.UUID      `json:"id"`
+	Type       string         `json:"type"`
+	ReleaseMsg ReleaseMessage `json:"template_data"`
```
CermakM, line 12: Calling it `template_data` in the db may not be the best wording, it's quite confusing.

service/model/template.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/google/uuid"
+)
+
+const (
+	releaseMsgTmplType = "release_message"
+)
+
+type TemplateRepository interface {
+	Insert(ctx context.Context, t Template) (Template, error)
```
CermakM, line 16: Use `Create` rather than `Insert`

service/model/release_message.go
CermakM, line 1: If I am not mistaken, this is part of the template. If it's not used elsewhere (outside of the enclosing Template model), then I would rather put it in the template file.

The same applies in the repository and transport layers.

service/model/template.go
```diff
@@ -0,0 +1,48 @@
...
+	ReadAll(ctx context.Context) ([]Template, error)
+	Read(ctx context.Context, id uuid.UUID) (Template, error)
+	Update(ctx context.Context, t Template) (Template, error)
+	Delete(ctx context.Context, id uuid.UUID) error
+}
+
+type Template struct {
```
CermakM, line 23: The template is not related to any user or project.

service/template.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/google/uuid"
+)
+
+type TemplateService struct {
+	repository model.TemplateRepository
+}
+
+func (s *TemplateService) Create(ctx context.Context, t model.Template) (model.Template, error) {
+	return s.repository.Insert(ctx, t)
+}
```
CermakM, line 17: The service should probably receive userID or orgID to only return templates that the user is authorized to access. Same applies to other endpoints. If I am not mistaken, the authorization is missing from all of them.

CermakM (changes requested):

jan-zabloudil:
We have decided not to include the templates feature in the MVP.
Instead, at the project level, users will be able to configure which information to include in their release messages.

PR #13, opened 2024-03-28
feat: add project endpoints

jan-zabloudil:
This pull request introduces the projects. The functionality for 'project invitations' and 'project members' will be added in separate pr.


service/model/project.go
```diff
@@ -0,0 +1,91 @@
...
+
+type ProjectRepository interface {
+	Create(ctx context.Context, p Project) (Project, error)
+	Read(ctx context.Context, id uuid.UUID) (Project, error)
+	ReadAll(ctx context.Context) ([]Project, error)
+	Update(ctx context.Context, p Project) (Project, error)
+	Delete(ctx context.Context, id uuid.UUID) error
+}
+
+type Project struct {
```
jan-zabloudil, line 20: Project contains `ReleaseAlertConfig` and `SlackConfig`.

`ReleaseAlertConfig` replaces the original concept of release message templates and determinates which information to include into release alert. I also decided to use `ReleaseAlert` instead of `ReleaseMessage`.

`SlackConfig` has only one field for now - channel id.

service/model/release_alert.go
jan-zabloudil, line 1: For now only `ReleaseAlertConfig` is only used as part of Project model.

But I decided to define `ReleaseAlert` already in order to better imagine how ReleaseAlertConfig will be used. And therefore I created separete file `/service/model/release_alert`.

`ReleaseAlert` holds the alert data. `ReleaseMessageBuilder` will be used for building the alert and will accept `ReleaseAlertConfig` as param.

service/project.go
jan-zabloudil, line 1: Project service functions do not have any authorization logic yet because project members feature will be implemented in the separate PR.

service/model/project_test.go
jan-zabloudil, line 1: I only added tests for service model functions.

Do you think it also make sense to mock repositories and write tests for service functions?

service/model/project.go
```diff
@@ -0,0 +1,91 @@
...
+	p.UpdatedAt = time.Now()
+
+	if err := p.Validate(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Project) Validate() error {
```
jan-zabloudil, line 81: Validate function enforces all business rules and is called when creating or updating project object. 

I prefer to not use any validator package in service layer.

repository/model/project.go
```diff
@@ -0,0 +1,94 @@
...
+		ID:                 id,
+		Name:               name,
+		Description:        description,
+		ReleaseAlertConfig: svcmodel.ReleaseAlertConfig(rlsCfg),
+		SlackConfig:        svcmodel.SlackConfig(slackCfg),
+		CreatedAt:          createdAt.Time,
+		UpdatedAt:          updatedAt.Time,
+	}
+
+	if err := p.Validate(); err != nil {
```
jan-zabloudil, line 77: When converting db object into service object I also perform validation

transport/model/project.go
jan-zabloudil, line 1: I prefer to have a separate struct for each operation, such as CreateProjectRequest, UpdateProjectRequest, and ProjectResponse.

In the transport layer, I only perform input validation to ensure that the user submits the correct payload. Any business rules are validated in the service layer

service/model/release_alert.go
jan-zabloudil, line 1: `ReleaseAlertConfig` is a part of the Project struct, currently not used outside this context. 

But to illustrate the potential use of Config, I already have created a `ReleaseAlert` struct and a builder for it. And therefore I put all the related code into `/service/model/release_alert`. The ReleaseAlert struct is designed to store alert data. The builder, which takes a Config as a parameter, builds the alert.

service/project.go
jan-zabloudil, line 1: There is no authorization logic in service functions for now because project members feature will be added in separate PR.

service/model/project_test.go
jan-zabloudil, line 1: I created tests only for service model functions for now.

Would you also mock repositories and write tests for service functions?

transport/utils/validate.go
```diff
@@ -20,6 +21,16 @@ func init() {
 	uni := ut.New(english, english)
 	translator, _ = uni.GetTranslator("en")
 	_ = entrans.RegisterDefaultTranslations(Validate, translator)
+
+	Validate.RegisterTagNameFunc(func(fld reflect.StructField) string {
```
bolshakoff, line 25: A bit confused about this change; especially because the `init()` function is not used anywhere.. 🤔

supabase/migrations/20240328123715_projects.sql
```diff
@@ -0,0 +1,27 @@
+create table "public"."projects" (
+                                     "id" uuid not null default gen_random_uuid(),
+                                     "name" text not null,
+                                     "description" text,
+                                     "release_alert_config" json,
+                                     "slack_config" json,
```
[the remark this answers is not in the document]
jan-zabloudil: I understand your point, but the issue is that Supabase does not support transactions, which complicates scenarios where you want to insert data into multiple tables. 
To achieve this, I would need to write a PostgreSQL function, which is not very maintainable. In this case, I feel that having the configuration saved in JSON is sufficient, but I will discuss it further with Marek during our call. Thank you!

supabase/migrations/20240328123715_projects.sql
```diff
@@ -0,0 +1,27 @@
...
+
+grant references on table "public"."projects" to "service_role";
+
+grant select on table "public"."projects" to "service_role";
+
+grant trigger on table "public"."projects" to "service_role";
+
+grant truncate on table "public"."projects" to "service_role";
+
+grant update on table "public"."projects" to "service_role";
```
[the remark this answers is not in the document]
jan-zabloudil: It was auto-generated migration. You are right, I fixed it

transport/utils/validate.go
jan-zabloudil: Should be called automatically when i import the package.

This change ensures that when validation errors occur, the field names in the error messages correspond to the JSON tags of the fields, which is more user-friendly.

jan-zabloudil:
This branch is now rebased onto https://github.com/jan-zabloudil/release-manager/pull/16 because during review I also rebased it to work with more recent code.

pkg/dberrors/dberrors.go
```diff
@@ -46,10 +46,9 @@ func NewMultipleOrNoRecordsReturnedError() *DBError {
 	}
 }
 
-func NewUnknownError(err error) *DBError {
+func NewUnknownError() *DBError {
 	return &DBError{
```
jan-zabloudil, line 50: Not related to this PR but addresses an issue left over from a previous PR.

service/model/user_role.go
```diff
@@ -7,22 +7,25 @@ import (
 type UserRole string
 
 const (
-	UserRoleAdmin UserRole = "admin"
-	UserRoleUser  UserRole = "user"
+	UserRoleAdmin   UserRole = "admin"
+	UserRoleManager UserRole = "manager"
+	UserRoleUser    UserRole = "user"
```
jan-zabloudil, line 12: New role `Manager` introduces. Why?

- Admin  role - have access to organization settings (implemented in #14) and can create projects as well

- Manager role - does not have access to organization settings, but can create projects _(role for Engineering managers in STRV)_

Other naming option would be `SuperAdmin` and `Admin roles`.

If you approve the addition of this new role, I will also create an endpoint that enables admin users to assign the manager role to other users.

service/project.go
```diff
@@ -0,0 +1,95 @@
...
+	p, err := model.NewProject(pc.Name, pc.SlackChannelID, pc.RlsAlertCfg)
+	if err != nil {
+		return model.Project{}, apierrors.NewProjectUnprocessableError().Wrap(err).WithMessage(err.Error())
+	}
+
+	return s.repository.Create(ctx, p)
+}
+
+func (s *ProjectService) Get(ctx context.Context, projectID uuid.UUID, authUserID uuid.UUID) (model.Project, error) {
+	// TODO add project member authorization
```
jan-zabloudil, line 39: Authorization will be added in the next PR, where project members feature will be added.

service/model/project.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+type ProjectUpdate struct {
+	ProjectID         uuid.UUID
+	Name              *string
+	SlackChannelID    *string
+	RlsAlertCfgUpdate ReleaseAlertConfigUpdate
+}
+
+type ReleaseAlertConfig struct {
```
jan-zabloudil, line 45: `ReleaseAlertConfig` defines the information that will be included in a release alerts for a given project. 

I chose 'ReleaseAlert' over 'ReleaseMessage' as I believe it’s a more suitable name

service/model/user_role.go
bolshakoff: Interesting point! I'd definitely steer towards something more simple and obvious, for example:
- org admin
- project admin
- user
How about other similar platforms? What names are they using?

And do we expect to have more different items?

If yes, consider simply having users and rights (read/modify org/project).

service/model/project.go
bolshakoff: `Alert` sounds too negative. :)
Also agree that `message` is a bit too neutral and general.
How about `notification`?

jan-zabloudil: Good point, let's go with `ReleaseNotification`

jan-zabloudil:
When discussing the release management app proposal, we considered two approaches for organizing releases:

**Project > Releases**
- originally proposed by @CermakM.
- each project represents a single application (e.g., Backend app for project X).
- one project = one GitHub repository.
- I suggest renaming it to `App > Releases`.
 
**Project > Apps > Releases**

I initially favored the second approach (Project > Apps > Releases) and built the entire proposal around this structure. But I have realized that it introduces unnecessary complexity without offering significant benefits. The first suggestion is much more flexible and aligns better with company workflows (including STRV).

**Why?**
Different teams typically handle different platforms for given project. For example, the Go team manages the backend solution, the iOS team handles the iOS app, and the Android team takes care of the Android app.

It makes sense for each Engineering Manager to create an app within ReleaseManager for which they are responsible. They can then invite engineers from their team with permissions to create releases. Managers can also invite engineers from other teams or PMs but only grant them read permissions. The manager will also connect the app to a GitHub repository and configure release notifications. 

But then, do we really need the project level at all? 
It seems that adding the project level would not bring any significant benefits and would only increase complexity. Adopting this approach would require managing permissions and settings on two levels. Or keep settings/permissions only on one level but significantly decrease the release manager app’s flexibility.

Since we are developing an open-source MVP, I would definitely choose the App > Releases solution.

pkg/apierrors/apierrors.go
```diff
@@ -54,7 +75,21 @@ func IsNotFoundError(err error) bool {
 	var apiErr *APIError
 	if errors.As(err, &apiErr) {
 		switch apiErr.Code {
-		case userNotFoundErrCode:
+		case userNotFoundErrCode, projectNotFoundErrCode:
```
CermakM, line 78: Just a suggestion here, but as you can see, the `IsError` functions have all the same definition and the only thing they differ in are the error codes comparisons. Meaning you can simplify this flow by implementing one `IsErrorWithCode` function of the signiture `IsErrorWithCode(err error, code Code)`.

pkg/responseerrors/responseerrors.go
```diff
@@ -9,14 +9,16 @@ import (
...
 	// #nosec G101 This is a constant error code, no security risk.
-	notBearerTokenFormatErrorCode = "ERR_TOKEN_NOT_BEARER_FORMAT"
-	missingBearerTokenErrCode     = "ERR_MISSING_BEARER_TOKEN"
-	defaultNotFoundErrCode        = "ERR_NOT_FOUND"
-	defaultForbiddenErrCode       = "ERR_FORBIDDEN"
-	defaultUnauthorizedErrCode    = "ERR_UNAUTHORIZED"
-	methodNotAllowedErrCode       = "ERR_METHOD_NOT_ALLOWED"
-	invalidResourceIDErrCode      = "ERR_INVALID_RESOURCE_ID"
-	unknownErrCode                = "ERR_UNKNOWN"
+	notBearerTokenFormatErrorCode     = "ERR_TOKEN_NOT_BEARER_FORMAT"
```
CermakM, line 12: I would rather structure the error codes as `errCodeX` rather than `xErrorCode`, this makes it easier to grep them both visually and via autocomplete.

repository/model/project.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/google/uuid"
+	"github.com/relvacode/iso8601"
+)
+
+type Project struct {
+	ID                 uuid.UUID          `json:"id"`
+	Name               string             `json:"name"`
+	SlackChannelID     string             `json:"slack_channel_id"`
+	ReleaseAlertConfig ReleaseAlertConfig `json:"release_alert_config"`
+	CreatedAt          iso8601.Time       `json:"created_at"`
```
CermakM, line 17: Why not `time.Time`?

repository/project.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func NewProjectRepository(c *supabase.Client) *ProjectRepository {
+	return &ProjectRepository{
+		client: c,
+		entity: "projects",
+	}
+}
+
+func (r *ProjectRepository) Create(ctx context.Context, p svcmodel.Project) (svcmodel.Project, error) {
```
CermakM, line 27: Is there any point in returning the same structure as passed as a parameter? To me it looks like this is an unnecessary overhead and simply returning an `error` would be enough.

service/model/project.go
```diff
@@ -0,0 +1,160 @@
...
+	ShowSourceCode  *bool
+}
+
+func NewProject(name, slackChannelID string, cfg ReleaseAlertConfig) (Project, error) {
+	p := Project{
+		ID:                 uuid.New(),
+		Name:               name,
+		SlackChannelID:     slackChannelID,
+		ReleaseAlertConfig: cfg,
+		CreatedAt:          time.Now(),
```
CermakM, line 71: Define a variable `now` and use it for both fields, otherwise they might not be equal, which is weird considering no update happened.

CermakM (changes requested):

CermakM:
Thanks for the elaboration on the entity structure! I 💯 agree 👍 

Regarding the user permissions:

> AppAdmin / ProjectAdmin
>
> Can create apps
  Does not have access to organization settings
  This role would typically be assigned to Engineering Managers
  
In the current STRV setting, I believe it's okay for us to have only organization admin and project member structure. However, from a long term perspective, I would absolutely consider having a project admin. Especially if the organization-wide settings grows in complexity and might potentially hold sensitive info. Also, since project administrators may not be technically savvy, there is some risk in allowing them access to org-wide settings.

repository/model/project.go
jan-zabloudil: Thanks! Initially, I had problems with marshaling/unmarshaling into a time.Time variable. But that was probably due to some beginner's mistakes. It works fine with `time.Time` for `timestamp with timezone` column 👍

jan-zabloudil:
Thank you 👍 I'm reassigning the PR to myself and will implement the remaining functionalities for `apps`.

jan-zabloudil:
- Environments have been added.

To keep PRs smaller, the following features will be implemented in separate PRs:
- Project members
- Connection to GitHub repository (it would be good to merge PR #14  first to have access to the GitHub token).

In the end, I also decided to continue referring to it as `Project` instead of `App`. It is more general, but I think it’s not necessarily a bad thing. Moreover, `project members` works better than `app members`, etc.

pkg/url/url.go
```diff
@@ -0,0 +1,14 @@
+package url
+
+import (
+	"net/url"
+)
+
+func IsAbsolute(raw string) bool {
+	u, err := url.Parse(raw)
+	if err != nil || u.Scheme == "" || u.Host == "" {
+		return false
```
jan-zabloudil, line 10: I'm not sure about this... I at least wanted to perform a basic check to see if a URL is valid according to a 'human' definition of a "correct" URL (of course, many incorrect ones would still pass). 

Or it might be better to use some existing validator...

jan-zabloudil:
Sorry, I've realized I need to implement a different approach to environments. Reassigning it to myself again and I'm on it.

pkg/url/url.go
bolshakoff: Yes, I think it can be done using the `validator` package.

jan-zabloudil:
> Sorry, I've realized I need to implement a different approach to environments. Reassigning it to myself again and I'm on it.

In STRV, all projects have development, staging, and production environments. However, I have decided not to have fixed environments to maintain flexibility, especially since I am developing an open source app.

1. When a manager creates a project, they can set up environments; the name field is required, while the URL and release fields are optional. _Potentially in STRV, we could automatically create development, staging, and production environments when a project is created_.

 2. When using the app, you can visit the environments page to see which releases are deployed to each environment, and you can quickly make changes.

Projects _ Environments

You can also "deploy" a release to an environment from the release page.

Projects _ Releases _ Add

pkg/url/url.go
jan-zabloudil: Thanks, I am going to change it in separate PR once this one is merged 👍

service/model/environment.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+type EnvironmentUpdate struct {
+	ID        uuid.UUID
+	ProjectID uuid.UUID
+	Name      *string
+	RawURL    *string
+}
+
+func NewEnvironment(projectID uuid.UUID, name, rawURL string) (Environment, error) {
```
CermakM, line 51: Why do I need to provide this URL? It should be accesible on the base url + id, shouldn't it?

service/environment.go
```diff
@@ -0,0 +1,121 @@
...
+		authSvc:    authSvc,
+		projectSvc: projectSvc,
+		repository: repo,
+	}
+}
+
+func (s *EnvironmentService) Create(ctx context.Context, ec model.EnvironmentCreation, authUserID uuid.UUID) (model.Environment, error) {
+	// TODO add project member authorization
+
+	_, err := s.projectSvc.Get(ctx, ec.ProjectID, authUserID)
```
CermakM, line 30: Shouldn't this be only the admin's privilage?

service/environment.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, err
+	}
+
+	return envs, nil
+}
+
+func (s *EnvironmentService) Delete(ctx context.Context, projectID, envID uuid.UUID, authUserID uuid.UUID) error {
+	// TODO add project member authorization
+
+	_, err := s.projectSvc.Get(ctx, projectID, authUserID)
```
CermakM, line 110: Same here, project admin only I would say.

supabase/migrations/20240417150340_environments.sql
```diff
@@ -0,0 +1,13 @@
+create table "public"."environments" (
+         "id" uuid not null default gen_random_uuid(),
+         "project_id" uuid not null references public.projects on delete cascade,
+         "name" text not null,
```
CermakM, line 4: Should be unique per project.

service/environment.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/google/uuid"
+)
+
+type EnvironmentService struct {
+	authSvc    model.AuthService
+	projectSvc model.ProjectService
+	repository model.EnvironmentRepository
+}
+
+func NewEnvironmentService(authSvc model.AuthService, projectSvc model.ProjectService, repo model.EnvironmentRepository) *EnvironmentService {
```
CermakM, line 19: I think it's an unnecessary overhead to have environment creation as a separate service. It's part of the project resource, doesn't have a seaparate resource endpoint, hence I would put it under the project service. Not that you don't have to have a service per resource either, rather structure it per domain.

CermakM (changes requested):

pkg/dberrors/dberrors.go
```diff
@@ -6,10 +6,10 @@ import (
 )
 
 var (
-	notFoundErrCode            = "ERR_DB_RESOURCE_NOT_FOUND"
-	cannotMapToSvcModelErrCode = "ERR_DB_CANNOT_MAP_TO_SVC_MODEL"
-	multipleOrNoRecordsErrCode = "ERR_DB_MULTIPLE_OR_NO_RECORDS_RETURNED"
-	unknownErrCode             = "ERR_DB_UNKNOWN"
+	errCodeNotFound            = "ERR_DB_RESOURCE_NOT_FOUND"
+	errCodeCannotMapToSvcModel = "ERR_DB_CANNOT_MAP_TO_SVC_MODEL"
+	errCodeMultipleOrNoRecords = "ERR_DB_MULTIPLE_OR_NO_RECORDS_RETURNED"
```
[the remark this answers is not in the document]
jan-zabloudil: In the end I dont need that error anymore, I removed it.

service/model/environment.go
jan-zabloudil: Renamed to `ServiceURL` to improve clarity.

service/project.go
```diff
@@ -0,0 +1,233 @@
...
+		return model.Environment{}, err
+	}
+
+	env, err := model.NewEnvironment(ec.ProjectID, ec.Name, ec.ServiceRawURL)
+	if err != nil {
+		return model.Environment{}, apierrors.NewEnvironmentUnprocessableError().Wrap(err).WithMessage(err.Error())
+	}
+
+	if err := s.validateEnvironmentNameUnique(ctx, env.ProjectID, env.Name); err != nil {
+		return model.Environment{}, err
```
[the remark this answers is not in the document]
bolshakoff: Sorry, consider this resolved..

But actually..
I'd return the duplicate name error here; ideally simplify it to return `(*bool, error)`.
A bit more code, but more obvious, more readable.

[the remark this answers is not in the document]
bolshakoff: I mean `s.validateEnvironmentNameUnique()`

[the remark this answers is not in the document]
jan-zabloudil: Thank you for review, all very good points!

Just here I chose to return `(bool, error)`.
I understand that by returning `(*bool, error)`, you can distinguish the following cases based on bool value:
- true - the name is unique
- false - the name is not unique
- nil - it cannot be determined if the name is unique

But in case of `(bool, error)` I simply check if err is nil. If it isn’t, I know that the uniqueness couldn't be determined due to an error. If err is nil, I can use the bool value directly without needing to dereference it.

service/model/environment.go
```diff
@@ -119,7 +110,7 @@ func (e *Environment) Update(name, rawURL *string) error {
 	}
 
 	if rawURL != nil {
-		if err := e.SetURL(*rawURL); err != nil {
+		if err := e.SetServiceURL(*rawURL); err != nil {
```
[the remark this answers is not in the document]
jan-zabloudil: > (optional) And actually, it might be even more clean to have the following signature:
> 
>     * `Update(u EnvironmentUpdate) (*Environment, error)`

Not sure if I get the point. What is the benefit of returning a pointer to the Environment in the case of a pointer receiver? Is it just for better readability?

```
env, err := env.Update(...)
```

CermakM (approved):

PR #14, opened 2024-03-28
 feat: add settings service 

jan-zabloudil:
This PR adds settings service.

### DIFF
--- a/repository/model/settings.go
+++ b/repository/model/settings.go
@@ -1,0 +1,113 @@
+package model
+
+import (
+	"encoding/json"
+	"fmt"
+
+	svcmodel "release-manager/service/model"
+)
+
+const (
+	keyOrganizationName      = "organization_name"
+	keyDefaultReleaseMessage = "default_release_message"
+	keySlack                 = "slack"
+	keyGithub                = "github"
+)
+
+type Settings struct {
+	OrganizationName      string
+	DefaultReleaseMessage string
+	Slack                 SlackSettings
+	Github                GithubSettings
+}
+
+type SlackSettings struct {
+	Enabled bool   `json:"enabled"`
+	Token   string `json:"token"`
+}
+
+type GithubSettings struct {
+	Enabled bool   `json:"enabled"`
+	Token   string `json:"token"`
+}
+
+func (s *Settings) MarshalJSON() ([]byte, error) {
+	type KeyValue struct {
+		Key   string `json:"key"`
+		Value any    `json:"value"`
+	}
+
+	return json.Marshal([]KeyValue{
+		{Key: keyOrganizationName, Value: s.OrganizationName},
+		{Key: keyDefaultReleaseMessage, Value: s.DefaultReleaseMessage},
+		{Key: keySlack, Value: s.Slack},
+		{Key: keyGithub, Value: s.Github},
+	})
+}
+
+func (s *Settings) UnmarshalJSON(data []byte) error {
+	type KeyRawValue struct {
+		Key   string          `json:"key"`
+		Value json.RawMessage `json:"value"`
+	}
+
+	var pairs []KeyRawValue
+	if err := json.Unmarshal(data, &pairs); err != nil {
+		return err
+	}
+
+	for _, kv := range pairs {
+		switch kv.Key {
+		case keyOrganizationName:
+			if err := json.Unmarshal(kv.Value, &s.OrganizationName); err != nil {
+				return fmt.Errorf("cannot unmarshal %s: %w", kv.Key, err)
+			}
+		case keyDefaultReleaseMessage:
+			if err := json.Unmarshal(kv.Value, &s.DefaultReleaseMessage); err != nil {
+				return fmt.Errorf("cannot unmarshal %s: %w", kv.Key, err)
+			}
+		case keySlack:
+			if err := json.Unmarshal(kv.Value, &s.Slack); err != nil {
+				return fmt.Errorf("cannot unmarshal %s: %w", kv.Key, err)
+			}
+		case keyGithub:
+			if err := json.Unmarshal(kv.Value, &s.Github); err != nil {
+				return fmt.Errorf("cannot unmarshal %s: %w", kv.Key, err)
+			}
+		default:
+			return fmt.Errorf("unknown key: %s", kv.Key)
+		}
+	}
+
+	return nil
+}
+
+func ToSettings(s svcmodel.Settings) Settings {
+	return Settings{
+		OrganizationName:      s.OrganizationName,
+		DefaultReleaseMessage: s.DefaultReleaseMessage,
+		Slack: SlackSettings{
+			Enabled: s.Slack.Enabled,
+			Token:   s.Slack.Token,
+		},
+		Github: GithubSettings{
+			Enabled: s.Github.Enabled,
+			Token:   s.Github.Token,
+		},
+	}
+}
+
+func ToSvcSettings(s Settings) svcmodel.Settings {
+	return svcmodel.Settings{
+		OrganizationName:      s.OrganizationName,
+		DefaultReleaseMessage: s.DefaultReleaseMessage,
+		Slack: svcmodel.SlackSettings{
+			Enabled: s.Slack.Enabled,
+			Token:   s.Slack.Token,
+		},
+		Github: svcmodel.GithubSettings{
+			Enabled: s.Github.Enabled,
+			Token:   s.Github.Token,
+		},
+	}
+}
--- a/service/model/settings.go
+++ b/service/model/settings.go
@@ -1,0 +1,122 @@
+package model
+
+import (
+	"errors"
+)
+
+var (
+	errOrganizationNameRequired  = errors.New("organization name is required")
+	errDefaultReleaseMsgRequired = errors.New("default release message is required")
+	errSlackMissingToken         = errors.New("token is required to enable slack integration")
+	errGithubMissingToken        = errors.New("token is required to enable github integration")
+)
+
+type Settings struct {
+	OrganizationName      string
+	DefaultReleaseMessage string
+	Slack                 SlackSettings
+	Github                GithubSettings
+}
+
+type UpdateSettingsInput struct {
+	OrganizationName  *string
+	DefaultReleaseMsg *string
+	Slack             UpdateSlackSettingsInput
+	Github            UpdateGithubSettingsInput
+}
+
+func (s *Settings) Update(u UpdateSettingsInput) error {
+	if u.OrganizationName != nil {
+		s.OrganizationName = *u.OrganizationName
+	}
+	if u.DefaultReleaseMsg != nil {
+		s.DefaultReleaseMessage = *u.DefaultReleaseMsg
+	}
+
+	if err := s.Slack.Update(u.Slack); err != nil {
+		return err
+	}
+	if err := s.Github.Update(u.Github); err != nil {
+		return err
+	}
+
+	return s.Validate()
+}
+
+func (s *Settings) Validate() error {
+	if s.OrganizationName == "" {
+		return errOrganizationNameRequired
+	}
+
+	if s.DefaultReleaseMessage == "" {
+		return errDefaultReleaseMsgRequired
+	}
+
+	if err := s.Slack.Validate(); err != nil {
+		return err
+	}
+
+	if err := s.Github.Validate(); err != nil {
+		return err
+	}
+
+	return nil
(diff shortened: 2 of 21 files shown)


repository/model/settings.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+const (
+	orgNameKey           = "organization_name"
+	slackTokenKey        = "slack_token"
+	githubTokenKey       = "github_access_token"
+	defaultReleaseMsgKey = "default_release_msg"
+)
+
+type Setting struct {
```
jan-zabloudil, line 18: In the database, settings are saved in a table with two columns: key and value. The value is stored as JSON. This approach enables easy addition of more setting options in the future

transport/transport.go
```diff
@@ -33,6 +35,9 @@ func NewHandler(us model.UserService) *Handler {
 		r.Delete("/", h.requireAdminUser(h.handleUser))
 	})
 
+	h.Mux.Patch("/organization/settings", h.requireAdminUser(h.updateSettings))
+	h.Mux.Get("/organization/settings", h.requireAdminUser(h.getSettings))
```
jan-zabloudil, line 39: These endpoints are available only to admin users. Authorization is performed by middleware in the transport layer. Or should the service layer also perform its own authorization?

jan-zabloudil: Not relevant anymore

pkg/autherrors/autherrors.go
```diff
@@ -34,10 +34,9 @@ func NewInvalidTokenError() *AuthError {
 	}
 }
 
-func NewUnknownError(err error) *AuthError {
+func NewUnknownError() *AuthError {
```
jan-zabloudil, line 37: Not related to the current PR but addresses an issue left over from a previous PR.

repository/model/setting.go
```diff
@@ -0,0 +1,109 @@
...
+	organizationNameKey  = "organization_name"
+	defaultReleaseMsgKey = "default_release_msg"
+	slackSettingKey      = "slack_setting"
+	githubSettingKey     = "github_setting"
+)
+
+type Setting struct {
+	Key     string `json:"key"`
+	Setting any    `json:"setting"`
+}
```
jan-zabloudil, line 20: Snímek obrazovky 2024-04-11 v 17 19 15

In the database, each setting is saved in JSON field. This approach simplifies the process of adding new settings.

jan-zabloudil:
Made some changes in `pkg/Xerrors` in https://github.com/jan-zabloudil/release-manager/pull/13.

I'm going to merge #13 first and then rebase this one before merging.

bolshakoff:
Waiting for rebase.. 👀

repository/model/config.go
```diff
@@ -0,0 +1,92 @@
...
+type IntegrationConfig struct {
+	Enabled bool   `json:"enabled"`
+	Token   string `json:"token"`
+}
+
+func (c *Config) MarshalJSON() ([]byte, error) {
+	type KeyValue struct {
+		Key   string `json:"key"`
+		Value any    `json:"value"`
+	}
```
jan-zabloudil, line 33: Config values are saved as key value pairs in db.

repository/config.go
```diff
@@ -0,0 +1,67 @@
...
+	data := model.ToConfig(
+		c.OrganizationName,
+		c.DefaultReleaseMessage,
+		c.Slack,
+		c.Github,
+	)
+
+	err := r.client.
+		DB.From(r.entity).
+		Upsert(&data).
```
jan-zabloudil, line 35: I've realized that it's more practical to pass data for updates and inserts as pointers, especially when a custom `MarshalJSON` function is implemented. To maintain consistency, I've applied this change to all insert and update operations.

`model.ToConfig()` could return a pointer directly, but I believe it is better to show the intention of passing pointer more explicitly.

supabase/seed.sql
jan-zabloudil, line 1: Default config is populated along with the database migrations.

transport/transport.go
```diff
@@ -60,6 +67,11 @@ func NewHandler(as model.AuthService, us model.UserService, ps model.ProjectServ
 		})
 	})
 
+	h.Mux.Route("/organization/config", func(r chi.Router) {
+		r.Get("/", h.requireAuthUser(h.getConfig))
+		r.Patch("/", h.requireAuthUser(h.updateConfig))
```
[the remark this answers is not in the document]
bolshakoff: Oof.. I think `Settings` was good; `Config` might be confusing, because it is typically associated with "technical" server config, etc.
And for example Jira, a similar software has `Settings`.

service/model/config.go
```diff
@@ -0,0 +1,139 @@
...
+		return errDefaultReleaseMsgRequired
+	}
+
+	return nil
+}
+
+func ToConfig(
+	orgName,
+	rlsMsg string,
+	slackEnabled bool,
```
bolshakoff, line 62: Also, a slippery place; I mentioned it a couple times previously, that it's better, simpler, safer to pass the whole type as the argument.
For example, here is again a danger of bug; because slack and github things have the same types, they can be messed up.

Also, I can see that it is used on the repo layer - I'd move it there.
It's fine for repo layer to import service layer types.

The general principle is that domain/service is the core of the application and everybody can import it; but domain/service layer doesn't really import anybody, and cares only about the business logic, and for example imports db via repository interface...

But I think I feel where you are coming from - this method signature feels more generic, more decoupled, so that repo layer doesn't have to import service layer.
I used to do the same, but soon realized, that passing the whole type is much more simple and safe.

service/model/config.go
```diff
@@ -0,0 +1,139 @@
...
+	slackToken string,
+	githubEnabled bool,
```
bolshakoff, line 64: In general, probably no need to do here, but in general, if the above comment taken to the extreme, you might want to wrap every input string into a custom type, i.e. in this case slackToken will have `SlackToken` type (which will simply embed `string`), etc.

Might sound extreme, but given the arguments above, might be worth it.
I took this idea from a conference from a cool Google guy or similar...

Yes, no need to do it here, because if you agree to wrap it into a type, it already will be much safer...

service/model/config.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	if err := c.Validate(); err != nil {
+		return IntegrationConfig{}, err
+	}
+
+	return c, nil
+}
+
+func (c *IntegrationConfig) Validate() error {
```
bolshakoff, line 129: Already now, I'd expect the validation of tokens to be different for different services.

service/model/config.go
```diff
@@ -0,0 +1,139 @@
...
+	if u.Enabled != nil {
+		c.Enabled = *u.Enabled
+	}
+	if u.Token != nil {
+		c.Token = *u.Token
+	}
+
+	// if integration is disabled, clear the token
+	if !c.Enabled {
+		c.Token = ""
```
bolshakoff, line 110: Is it documented? Will it be visible and expected by user?
If not, I wouldn't complicate it here and don't do anything...

bolshakoff (commented):
oh-oh

EDIT: duplicated review, disregard plz 🙏 

EDIT 2: Oh no, they refer to the same; so don't disregard pls 😄

repository/model/settings.go
```diff
@@ -0,0 +1,92 @@
...
+	Slack                 IntegrationSettings
+	Github                IntegrationSettings
+}
+
+type IntegrationSettings struct {
+	Enabled bool   `json:"enabled"`
+	Token   string `json:"token"`
+}
+
+func (c *Settings) MarshalJSON() ([]byte, error) {
```
bolshakoff, line 29: Why `c`?

repository/model/settings.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return json.Marshal([]KeyValue{
+		{Key: keyOrganizationName, Value: c.OrganizationName},
+		{Key: keyDefaultReleaseMessage, Value: c.DefaultReleaseMessage},
+		{Key: keySlack, Value: c.Slack},
+		{Key: keyGithub, Value: c.Github},
+	})
+}
+
+func (c *Settings) UnmarshalJSON(data []byte) error {
```
bolshakoff, line 43: It's not used anywhere.

repository/model/settings.go
bolshakoff: And it's not used anywhere.

bolshakoff:
Overall probably OK; the main two things is coupling of Slack and GitHub settings, and passing fields instead of whole structs.

Other than that, can't wait to see this in action. 😉

repository/model/settings.go
jan-zabloudil:
```
	err := r.client.
		DB.From(r.entity).
		Upsert(&data).
		ExecuteWithContext(ctx, nil)
```

When struct is passed to supabase client, client calls json.Marshal() and i need a custom marshal function to convert struct to key value pairs to save them in db.

repository/model/settings.go
jan-zabloudil: Same as above in case of `json.Unmarshal`

```
DB.From(r.entity).
Select("*").
ExecuteWithContext(ctx, &resp)
```

service/settings.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	settings, err := s.repository.Read(ctx)
+	if err != nil {
+		return model.Settings{}, err
+	}
+
+	if err := settings.Validate(); err != nil {
+		return model.Settings{}, apierrors.NewSettingsCorruptedError().Wrap(err)
+	}
```
jan-zabloudil, line 36: When settings is retrieved from repository, validate function is called in order to check settings validity.

jan-zabloudil:
Thanks for the review! Once this one is approved I am going to refactor other services and models in the spirit of suggested changes in thIs PR.

transport/model/config.go
```diff
@@ -0,0 +1,53 @@
...
+	Slack                 IntegrationConfigResponse `json:"slack"`
+	Github                IntegrationConfigResponse `json:"github"`
+}
+
+type IntegrationConfigResponse struct {
+	Enabled bool   `json:"enabled"`
+	Token   string `json:"token"`
+}
+
+func ToSvcConfigUpdate(orgName, rlsMsg *string, slack, github UpdateIntegrationConfigRequest) svcmodel.ConfigUpdate {
```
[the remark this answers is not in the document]
jan-zabloudil: Thanks, make sense 👍

One of the first reviews when I was passing the whole struct, Marek suggested to [pass each param explicitly](https://github.com/jan-zabloudil/release-manager/pull/4#discussion_r1517570203). But I probably got the point slightly wrong and therefore used this approach lately more widely than I should.

jan-zabloudil:
> Thanks for the review! Once this one is approved I am going to refactor other services and models in the spirit of suggested changes in thIs PR.

Before refactoring, I submitted one more PR to check if the chosen approach is correct -> #19

transport/model/config.go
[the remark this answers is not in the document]
bolshakoff: Aha, I see.
Also, good point by @CermakM - sort of mirror side effect of the "whole struct" approach..

But in general, I wouldn't be afraid of it, because you'll have it covered by tests, and if you miss something, you will notice and easily add it there.

It's fair to keep in mind though that we do not have end-to-end tests yet (as we discussed on Slack).
But still, I presume that it will be covered at least by some manual or semi-automatic testing at some point and will be discovered...

Also, looking at it a bit more - verbosity and semantics. This method does a trivial thing and should look trivial.
But when it has so many input arguments, it sort of hints that this method does something huge and that each argument plays an important role, when in reality it doesn't.

bolshakoff:
Omg, sorry for the duplicated reviews lately :D

service/model/settings.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+type Settings struct {
+	OrganizationName      string
+	DefaultReleaseMessage string
+	Slack                 SlackSettings
+	Github                GithubSettings
+}
+
+type SettingsUpdate struct {
```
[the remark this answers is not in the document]
jan-zabloudil: Ok, this one I am going to update in this PR. And then I will refactor the rest in the separate PR :)

What name would you use for the struct in the repository layer? Would you choose UpdateSettingsInput as well?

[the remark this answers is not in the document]
bolshakoff: On the repository layer, since you are upserting and always using the full type itself, there is no need to have a `UpdateSettingsInput` type, and that's totally fine.

PR #16, opened 2024-04-01
feat: add authentication and authorization service

jan-zabloudil:
This PR introduces several changes related to auth and users:

- Authentication and authorization were completely moved to the service layer. 
- Users are not read directly from `auth.users`; Instead, a `public users` table was added (as [recommended by Supabase](https://supabase.com/docs/guides/auth/managing-user-data)). This approach is more convenient because working directly with `auth.users` is not ideal; it cannot be used in joins (you need to write a PostgreSQL functions).
-  Tests have been added for the user and auth service.


repository/errors/auth.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/nedpals/supabase-go"
+)
+
+func ToSvcAuthErr(err error) error {
+	var supabaseErr *supabase.ErrorResponse
+	if errors.As(err, &supabaseErr) {
+		switch supabaseErr.Code {
+		case 401, 404:
+			return fmt.Errorf("%w: %s", svcerr.ErrAuthenticationFailed, err.Error())
```
CermakM, line 17: Here I guess you use the `NewAuthErr(message)` function mentioned in the previous comment.

repository/errors/user.go
```diff
@@ -0,0 +1,24 @@
...
+	"errors"
+	"fmt"
+
+	svcerr "release-manager/service/errors"
+
+	postgrestgo "github.com/nedpals/postgrest-go/pkg"
+)
+
+func ToSvcUserErr(err error) error {
+	var postgreErr *postgrestgo.RequestError
```
CermakM, line 13: Why error from this package?

transport/utils/context.go
CermakM, line 1: Singular name of the package.

repository/utils/utils.go
CermakM, line 1: Singular name.

transport/model/user.go
CermakM, line 1: Please, change the `ToNetUser` function names as well, I believe `ToUser` or `ToUserFromFields` should do well.

repository/error/error.go
jan-zabloudil, line 1: I see repository errors are causing lot of confusion. I simplified the whole logic.

In supabase client you can access:
- `Auth` that returns *supabase.ErrorResponse
- `DB` that returns postgres errors (*postgrestgo.RequestError)
- `Admin` (for working with auth.users) that return *supabase.ErrorResponse 

A separate function is defined to map errors for each case to service errors.

repository/error/error.go
```diff
@@ -0,0 +1,54 @@
...
+
+var (
+	ErrMultipleOrNoRecordsReturned = errors.New("supabase: one record expected, multiple (or no) rows returned")
+)
+
+func ToSvcErrFromAuth(err error) error {
+	var supabaseErr *supabase.ErrorResponse
+	if errors.As(err, &supabaseErr) {
+		switch supabaseErr.Code {
+		case 401, 404:
```
jan-zabloudil, line 24: In case of Auth - both 401, 404 errors means authentication failed (token is not valid or user does not exist anymore)

repository/error/error.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	return err
+}
+
+func ToSvceErrFromAdmin(err error) error {
+	var supabaseErr *supabase.ErrorResponse
+	if errors.As(err, &supabaseErr) {
+		switch supabaseErr.Code {
+		case 404:
```
jan-zabloudil, line 48: In Admin (which I use only for DELETE /admin/users/{id}) error 404 means that user was not found. I dont want to map any other error to service errors.

service/error/error.go
jan-zabloudil, line 1: I also simplified /service/error/error.go

I dont create specific error for each entity (such as `ErrUserNotFound`, `ErrProjectNotFound`). Instead, I have just  `ErrResourceNotFound`. Let me know if that is ok with you or if I should have specific error for each.

service/model/user.go
jan-zabloudil, line 1: Instead of having function `ToSvcUser` in `repository/model`. I added service function `ToUser` which is responsible for creating valid user from params. 

I use name `ToUser`, because `NewUser` would be used if function actually creates a new user. 

So the approach is basically:
-  You can create directly user with user role of any string (because user role is now `type UserRole string`). Implementation is not hidden behind interfaces.
- But service functions `NewUserRole` or `ToUser` are provided and you should use them in order to be sure valid objects are created.

jan-zabloudil:
Some changes suggested in this PR are relevant for other open PRs as well. I am going to update them once this PR is approved.

service/error/error.go
CermakM: It's often the case that you need to call multiple db calls in a single request, reading multiple resources. For example, assume you call products/{id} endpoint and get `ERR_RESOURCE_NOT_FOUND`, of course you assume it's the product that's missing. However, if the service layer has to perform multiple DB calls to retrieve the product, the client will never find out which of them was the issue.

service/model/user.go
CermakM: I like this approach 👍

repository/error/error.go
```diff
@@ -0,0 +1,54 @@
...
+		switch postgreErr.Code {
+		case PostgresResourceNotFoundCode:
+			return svcerr.NewResourceNotFoundErr(err.Error())
+		}
+	}
+
+	return err
+}
+
+func ToSvceErrFromAdmin(err error) error {
```
CermakM, line 44:
```suggestion
func ToSvcErrFromAdmin(err error) error {
```

repository/error/error.go
CermakM: I believe a better way would be to define a custom Error type for repository layer and one for API layer and then translate that type to a status code based on the code stored in the error. Meaning for example in the pkg/dberror, you would have an `Error` model that might look something like this:
```go
type Error struct {
    code string
    message string
    err error
}
```

The `dberrors` (note that there is 's' at the end, customary for errors) package could define just an `IsNotFound` (no specific resource needed) because the service layer would check if it's `IsNotFound` and translate that into for example `New<resource>NotFoundErr(...)` again with a custom code (a suitable place for such errors could be pkg/apierrors. Similarly the transport layer would translate the service error to a suitable status code.

CermakM (changes requested):

repository/user.go
```diff
@@ -44,6 +44,11 @@ func (r *UserRepository) Read(ctx context.Context, userID uuid.UUID) (svcmodel.U
 		resp.CreatedAt.Time,
 		resp.UpdatedAt.Time,
 	)
+	if err != nil {
+		return svcmodel.User{}, dberrors.NewToSvcModelErr(err)
+	}
```
jan-zabloudil, line 49:
```
u, err := svcmodel.ToUser(
...
)
```

Returns service error, but I want to return a repository error indicating that the database data could not be converted to the service model (with details provided by that service error).

service/svcerrors/user.go
```diff
@@ -0,0 +1,48 @@
...
+func NewUserErr(err error) error {
+	var repErr RepositoryError
+	if errors.As(err, &repErr) {
+		switch {
+		case repErr.IsNotFound():
+			return newUserNotFoundErr(err)
+		}
+	}
+
+	return newUserUnknownErr(err)
```
jan-zabloudil, line 28: If an error is not translated into a specific service error, I still want to return` <resource>UnknownError` to identify the service where the error occurred.

repository/dberrors/dberrors.go
```diff
@@ -0,0 +1,91 @@
...
+		switch supabaseErr.Code {
+		case 404:
+			return newResourceNotFoundErr(err)
+		}
+	}
+
+	return err
+}
+
+func NewDBErrFromAuth(err error) error {
```
CermakM, line 60: Does not return `DBError`

repository/dberrors/dberrors.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (e DBError) IsNotFound() bool {
+	return e.code == dbResourceNotFoundErrCode
+}
+
+func (e DBError) IsAuthInvalidToken() bool {
+	return e.code == authInvalidTokenErrCode
+}
+
+func NewDBErr(err error) error {
```
CermakM, line 40:
```suggestion
func NewDBError(err error) error {
```

service/svcerrors/svcerrors.go
```diff
@@ -0,0 +1,27 @@
+package svcerrors
+
+import "fmt"
+
+type RepositoryError interface {
```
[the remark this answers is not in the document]
CermakM: This is nice, that could work nicely!

repository/user.go
```diff
@@ -32,7 +34,15 @@ func (r *UserRepository) Read(ctx context.Context, userID uuid.UUID) (svcmodel.U
 		Eq("id", userID.String()).
 		ExecuteWithContext(ctx, &resp)
 	if err != nil {
-		return svcmodel.User{}, dberrors.NewDBErr(err)
+		var postgreErr *postgrestgo.RequestError
+		if errors.As(err, &postgreErr) {
+			switch postgreErr.Code {
+			case "PGRST116":
```
jan-zabloudil, line 40: Will be moved to helper function

transport/user.go
```diff
@@ -15,8 +15,20 @@ func (h *Handler) getUser(w http.ResponseWriter, r *http.Request) {
 		util.ContextAuthUserID(r),
 	)
 	if err != nil {
-		WriteApiErrorResponse(w, apierrors.NewApiError(err))
-		return
+		switch {
```
jan-zabloudil, line 18: Will be moved to helper function

transport/response.go
jan-zabloudil, line 1: I will keep only Write methods in transport/response.

And the rest will be put into `pkg/response_errors`

transport/util/errormapper.go
jan-zabloudil, line 1: I was struggling with the decision of where to place `ToResponseError`. Perhaps you have a better idea. thank you

CermakM (approved):

PR #17, opened 2024-04-06
chore: run golangci-lint in CI

jan-zabloudil:


CermakM:
Since you're at it, I would recommend adding linter as well: see [here](https://github.com/strvcom/strv-backend-go-net)

CermakM (approved):

PR #19, opened 2024-04-25
feat: add project invitations

jan-zabloudil:
Project members will be added in the separate PR.

### DIFF
--- a/repository/model/project_invitation.go
+++ b/repository/model/project_invitation.go
@@ -1,0 +1,69 @@
+package model
+
+import (
+	"time"
+
+	svcmodel "release-manager/service/model"
+
+	"github.com/google/uuid"
+)
+
+type ProjectInvitation struct {
+	ID            uuid.UUID `json:"id"`
+	ProjectID     uuid.UUID `json:"project_id"`
+	Email         string    `json:"email"`
+	ProjectRole   string    `json:"project_role"`
+	Status        string    `json:"status"`
+	TokenHash     []byte    `json:"token_hash"`
+	InviterUserID uuid.UUID `json:"invited_by"`
+	CreatedAt     time.Time `json:"created_at"`
+	UpdatedAt     time.Time `json:"updated_at"`
+}
+
+type ProjectInvitationUpdate struct {
+	Status    string    `json:"status"`
+	UpdatedAt time.Time `json:"updated_at"`
+}
+
+func ToProjectInvitation(i svcmodel.ProjectInvitation) ProjectInvitation {
+	return ProjectInvitation{
+		ID:            i.ID,
+		ProjectID:     i.ProjectID,
+		Email:         i.Email,
+		ProjectRole:   string(i.ProjectRole),
+		Status:        string(i.Status),
+		TokenHash:     i.TokenHash,
+		InviterUserID: i.InviterUserID,
+		CreatedAt:     i.CreatedAt,
+		UpdatedAt:     i.UpdatedAt,
+	}
+}
+
+func ToProjectInvitationUpdate(u svcmodel.ProjectInvitation) ProjectInvitationUpdate {
+	return ProjectInvitationUpdate{
+		Status:    string(u.Status),
+		UpdatedAt: u.UpdatedAt,
+	}
+}
+
+func ToSvcProjectInvitation(i ProjectInvitation) svcmodel.ProjectInvitation {
+	return svcmodel.ProjectInvitation{
+		ID:            i.ID,
+		ProjectID:     i.ProjectID,
+		Email:         i.Email,
+		ProjectRole:   svcmodel.ProjectRole(i.ProjectRole),
+		Status:        svcmodel.ProjectInvitationStatus(i.Status),
+		TokenHash:     i.TokenHash,
+		InviterUserID: i.InviterUserID,
+		CreatedAt:     i.CreatedAt,
+		UpdatedAt:     i.UpdatedAt,
+	}
+}
+
+func ToSvcProjectInvitations(invitations []ProjectInvitation) []svcmodel.ProjectInvitation {
+	i := make([]svcmodel.ProjectInvitation, 0, len(invitations))
+	for _, invitation := range invitations {
+		i = append(i, ToSvcProjectInvitation(invitation))
+	}
+	return i
+}
--- a/service/model/project_invitation.go
+++ b/service/model/project_invitation.go
@@ -1,0 +1,117 @@
+package model
+
+import (
+	"errors"
+	"time"
+
+	cryptox "release-manager/pkg/crypto"
+	validator "release-manager/pkg/validator"
+
+	"github.com/google/uuid"
+)
+
+const (
+	ProjectRoleEditor ProjectRole = "editor"
+	ProjectRoleViewer ProjectRole = "viewer"
+
+	InvitationStatusPending  ProjectInvitationStatus = "pending"
+	InvitationStatusAccepted ProjectInvitationStatus = "accepted_awaiting_registration"
+)
+
+var (
+	validProjectRoles = map[ProjectRole]bool{
+		ProjectRoleEditor: true,
+		ProjectRoleViewer: true,
+	}
+
+	validInvitationStatuses = map[ProjectInvitationStatus]bool{
+		InvitationStatusPending:  true,
+		InvitationStatusAccepted: true,
+	}
+
+	errProjectRoleInvalid             = errors.New("invalid project role")
+	errProjectInvitationStatusInvalid = errors.New("invalid invitation status")
+	errProjectInvitationEmailRequired = errors.New("email is required")
+	errProjectInvitationInvalidEmail  = errors.New("invalid email")
+)
+
+type ProjectInvitation struct {
+	ID            uuid.UUID
+	ProjectID     uuid.UUID
+	Email         string
+	ProjectRole   ProjectRole
+	Status        ProjectInvitationStatus
+	TokenHash     cryptox.Hash
+	InviterUserID uuid.UUID
+	CreatedAt     time.Time
+	UpdatedAt     time.Time
+}
+
+type CreateProjectInvitationInput struct {
+	ProjectID   uuid.UUID
+	Email       string
+	ProjectRole string
+}
+
+type ProjectRole string
+type ProjectInvitationStatus string
+
+func NewProjectInvitation(c CreateProjectInvitationInput, tkn cryptox.Token, inviterUserID uuid.UUID) (ProjectInvitation, error) {
+	now := time.Now()
+
+	i := ProjectInvitation{
+		ID:            uuid.New(),
+		ProjectID:     c.ProjectID,
+		Email:         c.Email,
+		ProjectRole:   ProjectRole(c.ProjectRole),
+		Status:        InvitationStatusPending,
+		TokenHash:     tkn.ToHash(),
+		InviterUserID: inviterUserID,
+		CreatedAt:     now,
+		UpdatedAt:     now,
+	}
+
+	if err := i.Validate(); err != nil {
+		return ProjectInvitation{}, err
+	}
+
+	return i, nil
+}
+
+func (i *ProjectInvitation) Accept() {
+	i.Status = InvitationStatusAccepted
+	i.UpdatedAt = time.Now()
+}
+
+func (i *ProjectInvitation) Validate() error {
+	if i.Email == "" {
+		return errProjectInvitationEmailRequired
+	}
+	if !validator.IsValidEmail(i.Email) {
+		return errProjectInvitationInvalidEmail
+	}
+	if err := i.ProjectRole.Validate(); err != nil {
+		return err
+	}
+	if err := i.Status.Validate(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r ProjectRole) Validate() error {
+	if _, exists := validProjectRoles[r]; exists {
+		return nil
+	}
+
+	return errProjectRoleInvalid
+}
+
+func (i ProjectInvitationStatus) Validate() error {
+	if _, exists := validInvitationStatuses[i]; exists {
+		return nil
+	}
+
+	return errProjectInvitationStatusInvalid
+}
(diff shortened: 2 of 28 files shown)


service/model/project_invitation.go
```diff
@@ -0,0 +1,126 @@
...
+	cryptox "release-manager/pkg/crypto"
+	validator "release-manager/pkg/validator"
+
+	"github.com/google/uuid"
+)
+
+const (
+	ProjectRoleOwner  ProjectRole = "owner"
+	ProjectRoleEditor ProjectRole = "editor"
+	ProjectRoleViewer ProjectRole = "viewer"
```
jan-zabloudil, line 16: Project role owner role will be assigned to the project creator.

service/model/project_invitation.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+const (
+	InvitationStatusPending  ProjectInvitationStatus = "pending"
+	InvitationStatusAccepted ProjectInvitationStatus = "accepted_awaiting_registration"
+)
```
jan-zabloudil, line 22: See `AcceptInvitation` function for explanation.

service/project.go
```diff
@@ -211,6 +219,147 @@ func (s *ProjectService) DeleteEnvironment(ctx context.Context, projectID, envID
...
+	if err != nil {
+		return nil, err
+	}
+
+	i, err := s.invitationRepo.ReadAllForProject(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := model.ValidateProjectInvitations(i); err != nil {
```
jan-zabloudil, line 273: Checking if invitations are returned in a valid state from the repository.

service/project.go
```diff
@@ -211,6 +219,147 @@ func (s *ProjectService) DeleteEnvironment(ctx context.Context, projectID, envID
...
+		return err
+	}
+
+	_, err := s.Get(ctx, projectID, authUserID)
+	if err != nil {
+		return err
+	}
+
+	i, err := s.invitationRepo.Read(ctx, invitationID)
+	if err != nil {
```
jan-zabloudil, line 291: When deleting an invitation, I don't check the invitation's validity.

service/project.go
```diff
@@ -211,6 +219,147 @@ func (s *ProjectService) DeleteEnvironment(ctx context.Context, projectID, envID
...
+		The function checks if the email is already associated with a user account.
+		If user exists, the function creates a project member and deletes the invitation.
+
+		When user signs up, function handle_new_user is triggered.
+		The function checks if the user's email is associated with a confirmed invitation(s).
+		If so, the function creates a project member(s) and deletes the invitation(s).
+
+		This approach is used because current repository implementation does not support transactions.
+	*/
+	invitation.Accept()
```
jan-zabloudil, line 320: The reason for `accepted_awaiting_registration` status.

Since Supabase doesn't support transactions and I don't control the user signing process via Google SSO, setting up database triggers is the best solution.

It will be implemented together with project members.

service/project.go
```diff
@@ -211,6 +219,147 @@ func (s *ProjectService) DeleteEnvironment(ctx context.Context, projectID, envID
...
+		return model.ProjectInvitation{}, apierrors.NewProjectInvitationCorruptedError().Wrap(err)
+	}
+
+	return i, nil
+}
+
+func (s *ProjectService) invitationExists(ctx context.Context, email string, projectID uuid.UUID) (bool, error) {
+	if _, err := s.invitationRepo.ReadByEmailForProject(ctx, email, projectID); err != nil {
+		if dberrors.IsNotFoundError(err) {
+			return false, nil
```
jan-zabloudil, line 354: Here, I am only checking if the invitation exists; therefore, no validation is necessary.

service/model/project_invitation.go
```diff
@@ -0,0 +1,119 @@
...
+	cryptox "release-manager/pkg/crypto"
+	validator "release-manager/pkg/validator"
+
+	"github.com/google/uuid"
+)
+
+const (
+	ProjectRoleEditor ProjectRole = "editor"
+	ProjectRoleViewer ProjectRole = "viewer"
+)
```
jan-zabloudil, line 16: Only two roles are necessary at the project level because all "admin" actions are granted only to admin users.

bolshakoff:
For such comments as above, feel free to put them into code. :)

service/project.go
bolshakoff: Aha, so that's because Supabase doesn't support transactions..

And could you explain more the Google SSO argument?

Other than that, probably OK, looks interesting.

Although, I'd say.. if you have time, I'd think whether not having transactions will introduce a really bad, inconsistent state.
If it's only about an extremely minimal chance of having a very unharmful state, then in the long run, consider moving everything to Go and re-ordering everything so that the number of really dangerous inconsistent states is minimized...

bolshakoff (changes requested):
Interesting.

service/project.go
```diff
@@ -211,6 +219,147 @@ func (s *ProjectService) DeleteEnvironment(ctx context.Context, projectID, envID
...
+		switch {
+		case dberrors.IsNotFoundError(err):
+			return model.ProjectInvitation{}, apierrors.NewProjectInvitationNotFoundError().Wrap(err)
+		default:
+			return model.ProjectInvitation{}, err
+		}
+	}
+
+	if err := i.Validate(); err != nil {
+		return model.ProjectInvitation{}, apierrors.NewProjectInvitationCorruptedError().Wrap(err)
```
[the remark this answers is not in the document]
bolshakoff: More sharp.

service/project.go
jan-zabloudil: Yes, it's true. Even my need to explain why I do validation in some places and not in others is not a good sign.

transport/model/interface.go
```diff
@@ -20,6 +21,12 @@ type ProjectService interface {
 	GetEnvironments(ctx context.Context, projectID, authUserID uuid.UUID) ([]svcmodel.Environment, error)
 	DeleteEnvironment(ctx context.Context, projectID, envID, authUserID uuid.UUID) error
 	UpdateEnvironment(ctx context.Context, u svcmodel.EnvironmentUpdate, projectID, envID, authUserID uuid.UUID) (svcmodel.Environment, error)
+
+	CreateInvitation(ctx context.Context, c svcmodel.ProjectInvitationCreation, authUserID uuid.UUID) (svcmodel.ProjectInvitation, error)
+	GetInvitations(ctx context.Context, projectID, authUserID uuid.UUID) ([]svcmodel.ProjectInvitation, error)
```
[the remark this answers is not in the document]
jan-zabloudil: I updated it for service relevant to this PR. Update for other services will be addressed in separate PR.

service/project.go
```diff
@@ -211,6 +219,147 @@ func (s *ProjectService) DeleteEnvironment(ctx context.Context, projectID, envID
 	return s.envRepo.Delete(ctx, envID)
 }
 
+func (s *ProjectService) CreateInvitation(ctx context.Context, c model.ProjectInvitationCreation, authUserID uuid.UUID) (model.ProjectInvitation, error) {
```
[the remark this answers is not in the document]
jan-zabloudil: > Since invitation has a separate table, separate models on all layers and a non-trivial logic, consider creating a separate service for it.

It's true that `ProjectService` is getting too big. I moved it to `ProjectMembershipService` - this service will manage project invitations and members.

> Or can there be another kind of invitation?

In the future we might add functionality to invite admin users to the app (not specific project). Currently, user has to sign up first and then admin role is assigned after. Which is not the best UX, but good enough for MVP.

Therefore I use route `/projects/invitations/accept|reject` (not just /invitations/..)

service/project.go
```diff
@@ -211,6 +219,147 @@ func (s *ProjectService) DeleteEnvironment(ctx context.Context, projectID, envID
...
+
+	return i, nil
+}
+
+func (s *ProjectService) GetInvitations(ctx context.Context, projectID, authUserID uuid.UUID) ([]model.ProjectInvitation, error) {
+	if err := s.authSvc.AuthorizeAdminRole(ctx, authUserID); err != nil {
+		return nil, err
+	}
+
+	_, err := s.Get(ctx, projectID, authUserID)
```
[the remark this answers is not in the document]
jan-zabloudil: Now it should be more clear - `s.projectSvc.Get`

service/project.go
jan-zabloudil: Another comment was, I will try to answer all of them it this thread 

> In general, I'm curious about db triggers, but..
>Apart from being detached from this, core place of business logic..
> Is it necessary?
> How will the debugging work?
> Do you check for example if invitation already has status accepted_... or other potentially invalid states?..

jan-zabloudil: > And could you explain more the Google SSO argument?

When a user signs up via Google SSO using Supabase, my backend API plays no role (it is not called at all). Therefore, the only way I can perform some action upon user registration is to set up a trigger in the database.

> Although, I'd say.. if you have time, I'd think whether not having transactions will introduce a really bad, inconsistent state.

If I move all logic to Go, I would:

1. Check if the user already has an account (1st query).

-> User has account
2. Delete the invitation (2nd query).
3. Create a project member (3rd query).

-> User has no account yet
2. Update invitation status (2nd query)

If something fails along the way (or changes), I would end up in an inconsistent state. 

>  Do you check for example if invitation already has status accepted_... or other potentially invalid states?..

Yes, I would check all these cases in a PostgreSQL function.

---

I agree with you that a downside of the solution is its testability and the fact that such core business logic is handled only in the database layer.

One of the objectives of my thesis is to evaluate the efficiency of using Supabase (and its REST API for database access instead of direct database connections). And I will mention all those downsides.

I have already spent quite lot of time debugging unofficial Supabase client for Go (there is no official yet). If this evaluation was not a part of my thesis goals, I would consider connecting to db directly.

transport/model/settings.go
```diff
@@ -38,7 +38,7 @@ type GithubSettings struct {
 	Token   string `json:"token"`
 }
 
-func ToSvcSettingsUpdate(u UpdateSettingsInput) svcmodel.UpdateSettingsInput {
+func ToSvcUpdateSettingsInput(u UpdateSettingsInput) svcmodel.UpdateSettingsInput {
```
jan-zabloudil, line 41: Just one leftover from previous the PR.

repository/model/project_invitation.go
```diff
@@ -0,0 +1,69 @@
...
+	Email         string    `json:"email"`
+	ProjectRole   string    `json:"project_role"`
+	Status        string    `json:"status"`
+	TokenHash     []byte    `json:"token_hash"`
+	InviterUserID uuid.UUID `json:"invited_by"`
+	CreatedAt     time.Time `json:"created_at"`
+	UpdatedAt     time.Time `json:"updated_at"`
+}
+
+type ProjectInvitationUpdate struct {
```
jan-zabloudil, line 23: In the repository layer, I believe `ProjectInvitationUpdate` fits better than `UpdateProjectInvitationInput`.

service/project.go
jan-zabloudil: Alternatively, instead of setting up a trigger, I can directly call a PostgreSQL function from the repository.

bolshakoff: Aha, OK, makes sense.

But still:
>If something fails along the way (or changes), I would end up in an inconsistent state.

What _exactly_ can go wrong?

For example if the account exists:
1. The account existence query is  (especially since it's such a rare, non-critical case in a non-real-time application).
2. You can rather first create an account, and then delete an invitation, in one service call. What can go wrong - during or after the project member creation the user will accept the invitation again? Weird, non-critical corner case, but you have a uniqueness constraint on `(project_id, email)`, so it will fail, you'll just have to handle this uniqueness violation error and wrap it into smth like `REPEATED_INVITATION_ACC_ERROR`...

Or it might be interesting to explore whether we can have even more consistency/integrity imposed declaratively, i.e. some references between the tables, maybe a new table, uniqueness constraints...

But fair enough, for research purposes, indeed, it might be interesting to see how helpful or dragging the db functions/triggers approach will be.

>Alternatively, instead of setting up a trigger, I can directly call a PostgreSQL function from the repository.

Yes, that sounds like a good compromise.
Although, you already have a comment there, so I guess we can keep it as it is.

service/project.go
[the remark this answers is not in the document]
bolshakoff: Yes, it's better now, but I'd expect simply a call like `Exists()`.
Or, if we want to re-use the `Get()` method, I'd definitely expect a clear error handling like:
`if err = dbNotFound { return ERR_PROJECT_NOT_FOUND }`

service/project.go
jan-zabloudil: Ok, thanks for comments! Anyway I am going to implement it in the next PR so I will consider all the options again. 👍

service/model/project_invitation.go
```diff
@@ -0,0 +1,119 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r ProjectRole) Validate() error {
+	switch r {
+	case ProjectRoleEditor, ProjectRoleViewer:
```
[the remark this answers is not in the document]
bolshakoff: Yes, I think I saw a similar pattern in some respectable library. 👍

repository/model/project_invitation.go
bolshakoff: Maybe. If I commented on this and voted for `_Input` name, it was a bit on the edge, so OK. 👍

bolshakoff (approved):
I think enough for this PR.
Changes are good, also good momentum in general.

Just one last actionable item, and a couple of general considerations, especially about db triggers and inconsistent states.

Without the constant validations and `CORRUPTED` errors, the code reads much smoother and trustworthy. 👌

bolshakoff:
Actually, one more thing, quite important - try to keep the changes and commits more small and atomic.
Especially when there are specific review comments, it should be especially easy to just address them comment by comment, commit by commit.
It makes verifying for reviewer (at least for me) much easier.

Also for the case if I suggest something stupid and @CermakM will come and ask to roll it back. 🤣 🙈 
(Would be one click "Revert commit", as opposed to having one huge "review" commit and rolling back changes manually. 😩)

jan-zabloudil:
Ok, I will keep that in mind, thank you 👍 and thanks for all the reviews!

PR #20, opened 2024-04-26
refactor: update users to be consistent with other domains

jan-zabloudil:
- Migration file updated
- Implemented a different approach to passing structs between layers implemented (whole structs instead of params)


repository/util/util.go
```diff
@@ -27,13 +27,6 @@ func ToDBError(err error) *dberrors.DBError {
 		}
 	}
 
-	var supabaseErr *supabase.ErrorResponse
-	if errors.As(err, &supabaseErr) {
-		if supabaseErr.Code == supabaseNotFoundErrorCode {
```
jan-zabloudil, line 32: Not needed anymore

service/model/user.go
```diff
@@ -6,6 +6,21 @@ import (
...
+	userRoleUserPriority  int = 2
+)
+
+var userRolePriority = map[UserRole]int{
+	UserRoleAdmin: userRoleAdminPriority,
+	UserRoleUser:  userRoleUserPriority,
+}
+
+type UserRole string
+
```
jan-zabloudil, line 23: `UserRole` moved to `/service/model/user.go`

supabase/migrations/20240225150436_custom_claims.sql
jan-zabloudil, line 1: Not needed anymore because we have public users table.

jan-zabloudil, line 1: Not needed anymore because we have `public.users` table and there is no need to modify `auth.users`.

supabase/migrations/20240331171034_public_users.sql
jan-zabloudil, line 1: Just code styling and enum for user role.

bolshakoff (approved):
Good

service/user.go
```diff
@@ -45,19 +45,15 @@ func (s *UserService) Delete(ctx context.Context, id uuid.UUID, authUserID uuid.
 		return err
 	}
 
-	if err := s.repository.Delete(ctx, id); err != nil {
```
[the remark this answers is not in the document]
jan-zabloudil: Yes, I just wanted to have it more consistent with the other domains. Because when deleting from `auth.users`, supabase provides endpoint for it and returns 404 if user does not exist. However, this is not the case when deleting from 'normal' tables, where I must first call a read function to check if the record exists so I can return a 404 error if it does not.

PR #21, opened 2024-04-26
refactor: update projects to be consistent with other domains

jan-zabloudil:
- Migration file updated
- Implemented a different approach to passing structs between layers implemented (whole structs instead of params)

### DIFF
--- a/service/model/environment.go
+++ b/service/model/environment.go
@@ -27,18 +27,18 @@ type Environment struct {
 	// TODO add release id once releases are implemented
 }
 
-type EnvironmentCreation struct {
+type CreateEnvironmentInput struct {
 	ProjectID     uuid.UUID
 	Name          string
 	ServiceRawURL string
 }
 
-type EnvironmentUpdate struct {
+type UpdateEnvironmentInput struct {
 	Name          *string
 	ServiceRawURL *string
 }
 
-func NewEnvironment(c EnvironmentCreation) (Environment, error) {
+func NewEnvironment(c CreateEnvironmentInput) (Environment, error) {
 	u, err := toServiceURL(c.ServiceRawURL)
 	if err != nil {
 		return Environment{}, err
@@ -66,32 +66,14 @@ func (e *Environment) Validate() error {
 		return errEnvironmentNameRequired
 	}
 
+	if e.ServiceURL.String() != "" && !validator.IsAbsoluteURL(e.ServiceURL.String()) {
+		return errEnvironmentServiceURLMustBeAbsolute
+	}
+
 	return nil
 }
 
-func ToEnvironment(id, projectID uuid.UUID, name, url string, createdAt, updatedAt time.Time) (Environment, error) {
-	u, err := toServiceURL(url)
-	if err != nil {
-		return Environment{}, err
-	}
-
-	env := Environment{
-		ID:         id,
-		ProjectID:  projectID,
-		Name:       name,
-		ServiceURL: u,
-		CreatedAt:  createdAt,
-		UpdatedAt:  updatedAt,
-	}
-
-	if err := env.Validate(); err != nil {
-		return Environment{}, err
-	}
-
-	return env, nil
-}
-
-func (e *Environment) Update(u EnvironmentUpdate) error {
+func (e *Environment) Update(u UpdateEnvironmentInput) error {
 	if u.ServiceRawURL != nil {
 		svcURL, err := toServiceURL(*u.ServiceRawURL)
 		if err != nil {
@@ -118,10 +100,6 @@ func toServiceURL(rawURL string) (url.URL, error) {
 	u, err := url.Parse(rawURL)
 	if err != nil {
 		return url.URL{}, errEnvironmentInvalidServiceURL
-	}
-
-	if !validator.IsAbsoluteURL(rawURL) {
-		return url.URL{}, errEnvironmentServiceURLMustBeAbsolute
 	}
 
 	return *u, nil
--- a/service/project.go
+++ b/service/project.go
@@ -31,7 +31,7 @@ func NewProjectService(
 	}
 }
 
-func (s *ProjectService) Create(ctx context.Context, c model.ProjectCreation, authUserID uuid.UUID) (model.Project, error) {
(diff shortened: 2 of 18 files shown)


service/model/environment.go
```diff
@@ -66,32 +66,14 @@ func (e *Environment) Validate() error {
...
-
-	env := Environment{
-		ID:         id,
-		ProjectID:  projectID,
-		Name:       name,
-		ServiceURL: u,
-		CreatedAt:  createdAt,
-		UpdatedAt:  updatedAt,
+	if e.ServiceURL.String() != "" && !urlx.IsAbsolute(e.ServiceURL.String()) {
+		return errEnvironmentServiceURLMustBeAbsolute
```
jan-zabloudil, line 70: I am going to update it to use validator package in the separate PR.

service/project_membership.go
```diff
@@ -38,8 +38,10 @@ func (s *ProjectMembershipService) CreateInvitation(
 		return model.ProjectInvitation{}, err
 	}
 
-	if _, err := s.projectSvc.Get(ctx, c.ProjectID, authUserID); err != nil {
+	if exists, err := s.projectSvc.Exists(ctx, c.ProjectID, authUserID); err != nil {
 		return model.ProjectInvitation{}, err
+	} else if !exists {
+		return model.ProjectInvitation{}, apierrors.NewProjectNotFoundError()
 	}
```
jan-zabloudil, line 45: Clearer error handling

service/project.go
```diff
@@ -134,9 +152,10 @@ func (s *ProjectService) CreateEnvironment(ctx context.Context, c model.Environm
 func (s *ProjectService) GetEnvironment(ctx context.Context, projectID, envID, authUserID uuid.UUID) (model.Environment, error) {
 	// TODO add project member authorization
 
-	_, err := s.Get(ctx, projectID, authUserID)
-	if err != nil {
+	if exists, err := s.Exists(ctx, projectID, authUserID); err != nil {
```
bolshakoff, line 155: Doesn't read smoothly - we are in an `Environment`-related method, and query `Exists()` and it's about a project.

Would expect either `ProjectExists()` or maybe also move environment-related operations to a separate service, similar to what we did with `project_membership`.

Deserves a thought, mainly up to you, but from my side a simple argument, a simple rule of thumb - environment has a separate fully-fledged model with creation and update counterparts, so it's at least CRUD, which seems significant enough to indeed separate it.

Another argument is obviously the expectation how will they grow (project and environment domain).

Also, from a usage / reader perspective, again, there are now for example two methods in this service:
- `Create()` - creates project
- `CreateEnvironment()` - creates environment for a project

Doesn't feel smooth...

service/project.go
```diff
@@ -134,9 +152,10 @@ func (s *ProjectService) CreateEnvironment(ctx context.Context, c model.Environm
 func (s *ProjectService) GetEnvironment(ctx context.Context, projectID, envID, authUserID uuid.UUID) (model.Environment, error) {
```
bolshakoff, line 152: Since we are here, another thing that constantly irritates my eye, albeit only slightly, is - why do we return zero struct when the environment does not exist?
Wouldn't it be simpler to have a pointer there?
It would also allow for simpler checks in the transport layer, if it decides (yes, we return `not exists error` also, but often you just want to be able to do simple `if env == nil ...`.

I personally in my experience never had problems with pointer approach.

Applicable to the rest of domains obviously.

service/project.go
```diff
@@ -76,15 +92,16 @@ func (s *ProjectService) Delete(ctx context.Context, projectID uuid.UUID, authUs
 		return err
 	}
 
-	_, err := s.Get(ctx, projectID, authUserID)
-	if err != nil {
+	if exists, err := s.Exists(ctx, projectID, authUserID); err != nil {
 		return err
+	} else if !exists {
```
bolshakoff, line 97: Still, there's some friction when reading this. :D

An `if` statement ending not with `else`. Meh.. 😕 

Consider flattening it a bit. I don't mind additional lines of code.

service/project.go
```diff
@@ -76,15 +92,16 @@ func (s *ProjectService) Delete(ctx context.Context, projectID uuid.UUID, authUs
 		return err
 	}
 
-	_, err := s.Get(ctx, projectID, authUserID)
-	if err != nil {
+	if exists, err := s.Exists(ctx, projectID, authUserID); err != nil {
 		return err
```
bolshakoff, line 96: Since we are in a refactoring / cleanup phase, see if you can send an error from here in a following fashion:
`return fmt.Errorf("checking if project exists: %w", err)`

service/project.go
bolshakoff: flattening, making it more simple and obvious...

bolshakoff (commented):
Nice. Very nice.

service/project_membership.go
bolshakoff: Yes 👍 
But still, not 100% smooth 😄 
See comment below 😉

service/model/environment.go
bolshakoff: Yeah, already approved. 💪

service/project.go
bolshakoff: Might be a separate PR of course

service/project.go
bolshakoff: Also probably a separate PR (if you agree)

service/project.go
bolshakoff: Probably a separate PR, if we decide to invest in better "erroring".

service/project.go
jan-zabloudil: I created an issue #23. We can discuss it there without blocking this PR. Also I am going to drop the second commit that ads `exists` method for now.

service/project.go
jan-zabloudil: Created issue for it #24

service/project.go
jan-zabloudil: I am going to drop the second commit, and we can discuss the best solution in issue #23, as this seems to have been the catalyst for discussions about moving the environment to a separate service

service/project.go
jan-zabloudil: Created #25

bolshakoff (approved):

PR #26, opened 2024-04-27
feat: enable to set github repository for a project

jan-zabloudil:
To keep PRs smaller, this one only adds the possibility to set a GitHub repo for the project. Accessing the GitHub API (using the [GitHub Go client](https://github.com/google/go-github)) will be handled in upcoming PRs
 
Originally, we wanted to create a separate endpoint for setting a project's GitHub repository. However, I've realized that it isn’t necessary and that it would be much simpler to handle this as part of the project creation and update process.

I was also considering implementing some sort of abstraction. Instead of having a `GithubRepository`, we could use something like `SCMRepository` in case we decide to add GitLab integration or other platforms. However, I think it’s better to keep it simple for now and then refactor it once it’s necessary and we have all the information needed to do the abstraction correctly.

### DIFF
--- a/service/model/project.go
+++ b/service/model/project.go
@@ -2,13 +2,27 @@ package model
 
 import (
 	"errors"
+	"net/url"
+	"strings"
 	"time"
+
+	"release-manager/pkg/validator"
 
 	"github.com/google/uuid"
 )
 
+const (
+	// expectedGithubRepositoryURLSlugCount is the expected number of slugs in a GitHub repository URL
+	// Example URL: https://github.com/owner/repo -> owner and repo are the slugs
+	expectedGithubRepositoryURLSlugCount = 2
+
+	githubHost    = "github.com"
+	githubHostWWW = "www.github.com"
+)
+
 var (
-	errProjectNameRequired = errors.New("project name is required")
+	errProjectNameRequired               = errors.New("project name is required")
+	errProjectInvalidGithubRepositoryURL = errors.New("invalid GitHub repository URL")
 )
 
 type Project struct {
@@ -16,6 +30,7 @@ type Project struct {
 	Name                      string
 	SlackChannelID            string
 	ReleaseNotificationConfig ReleaseNotificationConfig
+	GithubRepository          GithubRepository
 	CreatedAt                 time.Time
 	UpdatedAt                 time.Time
 }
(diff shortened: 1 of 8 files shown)


service/model/project.go
```diff
@@ -82,30 +123,6 @@ func (p *Project) Update(u UpdateProjectInput) error {
 	return p.Validate()
 }
 
-func ToProject(
```
jan-zabloudil, line 85: Just leftover from previous PR. This function is not used anymore.

service/model/project.go
```diff
@@ -138,3 +155,45 @@ func (c *ReleaseNotificationConfig) Update(u UpdateReleaseNotificationConfigInpu
...
+	if rawURL == "" {
+		return GithubRepository{}, nil
+	}
+
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return GithubRepository{}, errProjectInvalidGithubRepositoryURL
+	}
+
+	if !validator.IsAbsoluteURL(rawURL) {
```
CermakM, line 171: Probably move before parsing of the URL.

service/model/project.go
```diff
@@ -138,3 +155,45 @@ func (c *ReleaseNotificationConfig) Update(u UpdateReleaseNotificationConfigInpu
...
+		return GithubRepository{}, errProjectInvalidGithubRepositoryURL
+	}
+
+	if u.Host != githubHost && u.Host != githubHostWWW {
```
CermakM, line 175: Isn't this a bit of an excessive check? If they want to break it, they can do that anyway by providing a URL to a non-existing repository, just starting with `github.com`.

service/model/project.go
```diff
@@ -138,3 +155,45 @@ func (c *ReleaseNotificationConfig) Update(u UpdateReleaseNotificationConfigInpu
...
+		return GithubRepository{}, errProjectInvalidGithubRepositoryURL
+	}
+
+	/*
```
CermakM, line 179: No need for the comment, the logic is clear.

CermakM (approved):
In general, LGTM 👍 Just a couple of comments that you may ignore if you wish.

service/model/project.go
jan-zabloudil: Thank you, that's true. If the check was more complicated, I would definitely remove it, as it isn't worth it since it can be easily broken.

But it's a very simple check, it can still effectively handle basic cases when someone might accidentally provide the wrong URL (not intentionally). And it is already there so I keep it for now.

PR #28, opened 2024-04-28
feat: add project member models

jan-zabloudil:
I will also add model tests if you agree with this structure.

### DIFF
--- a/repository/model/project_member.go
+++ b/repository/model/project_member.go
@@ -1,0 +1,23 @@
+package model
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type CreateProjectMemberInput struct {
+	UserID      uuid.UUID `json:"user_id"`
+	ProjectID   uuid.UUID `json:"project_id"`
+	ProjectRole string    `json:"project_role"`
+	CreatedAt   time.Time `json:"created_at"`
+	UpdatedAt   time.Time `json:"updated_at"`
+}
+
+type ProjectMember struct {
+	User        User      `json:"users"` // Supabase returns joined table data in json array named after joined table, "users" in this case
+	ProjectID   uuid.UUID `json:"project_id"`
+	ProjectRole string    `json:"project_role"`
+	CreatedAt   time.Time `json:"created_at"`
+	UpdatedAt   time.Time `json:"updated_at"`
+}
--- a/service/model/project_membership.go
+++ b/service/model/project_membership.go
@@ -11,8 +11,13 @@ import (
 )
 
 const (
+	ProjectRoleOwner  ProjectRole = "owner"
 	ProjectRoleEditor ProjectRole = "editor"
 	ProjectRoleViewer ProjectRole = "viewer"
+
+	projectRoleOwnerPriority  int = 1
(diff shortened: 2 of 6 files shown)


service/model/project_membership.go
```diff
@@ -115,3 +131,64 @@ func (i ProjectInvitationStatus) Validate() error {
...
 	return errProjectInvitationStatusInvalid
 }
+
+func (r ProjectRole) IsRoleAtLeast(role ProjectRole) bool {
+	return projectRolePriority[r] <= projectRolePriority[role]
+}
+
+type ProjectMember struct {
+	User        User
+	ProjectID   uuid.UUID
```
jan-zabloudil, line 141: (@bolshakoff in the end I decided to handle project member creation in Go.)

I went back and forth on this, but I think it's good to include `User` as part of the `ProjectMember` struct. In most cases, I will need at least the user’s name and email every time I work with a project member. And knowing the user's role might be useful, as it can override the project role in the case of an admin user role.

service/model/project_membership.go
```diff
@@ -11,28 +11,41 @@ import (
 )
 
 const (
+	ProjectRoleOwner  ProjectRole = "owner"
```
jan-zabloudil, line 14: I added the project role `owner`, even though it is not currently needed for granting privileges since all project admin privileges are based on the user role admin. 

However, it is still useful to identify who created the project. We might extend privileges in the future, and having this role will be useful.

repository/model/project_member.go
```diff
@@ -0,0 +1,23 @@
+package model
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type CreateProjectMemberInput struct {
```
jan-zabloudil, line 9: @bolshakoff I think this naming feels appropriate for the repository layer after all. I'll update it in the other models as well.

transport/model/project_member.go
```diff
@@ -0,0 +1,16 @@
+package model
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type ProjectMember struct {
```
jan-zabloudil, line 9: The response in the transport layer can be simplified to look like this. It returns all the information needed regarding the project member.

CermakM (approved):

jan-zabloudil:
I'm going to add tests and then merge it.

service/model/project_membership.go
```diff
@@ -11,28 +11,41 @@ import (
...
+	projectRoleEditorPriority int = 2
+	projectRoleViewerPriority int = 3
+
 	InvitationStatusPending  ProjectInvitationStatus = "pending"
 	InvitationStatusAccepted ProjectInvitationStatus = "accepted_awaiting_registration"
 )
 
 var (
 	validProjectRoles = map[ProjectRole]bool{
+		ProjectRoleOwner:  true,
```
bolshakoff, line 28: Is it also included on the db layer?

bolshakoff (approved):

service/model/project_membership.go
jan-zabloudil: Thanks! I forgot, I'll add it.

Issue #25, opened 2024-04-26
Consider error handling improvements


jan-zabloudil:
https://github.com/jan-zabloudil/release-manager/pull/21#discussion_r1581128376

https://github.com/jan-zabloudil/release-manager/pull/29#discussion_r1583100401

https://github.com/jan-zabloudil/release-manager/pull/29#discussion_r1583122465

https://github.com/jan-zabloudil/release-manager/pull/29#discussion_r1583129829

bolshakoff:
I guess you'd arrive at it naturally when debugging the application, but it's good that we make it explicit and maybe even proactive here. 🤙

bolshakoff:
Also, I mentioned a couple times that having errors defined outside the domains / 3rd party services packages doesn't always feel smooth.

I guess can be discussed here.

jan-zabloudil:
Done 🎉

PR #29, opened 2024-04-28
feat: retrieve git tags from a GitHub repository

jan-zabloudil:


service/model/project.go
```diff
@@ -193,3 +197,7 @@ func toGithubRepository(rawURL string) (GithubRepository, error) {
 		RepositorySlug: slugs[1],
 	}, nil
 }
+
+type GitTag struct {
+	Name string
+}
```
jan-zabloudil, line 203: GitHub API returns only the git tag name, but this is sufficient. From the user's perspective, they will simply select a tag to link the release to the source code. When creating a GitHub release, GitHub requires only the git tag name, so no additional parameters are needed

service/project.go
```diff
@@ -2,32 +2,45 @@ package service
...
-	authSvc        model.AuthService
-	projectRepo    model.ProjectRepository
-	envRepo        model.EnvironmentRepository
-	invitationRepo model.ProjectInvitationRepository
+	authSvc             model.AuthService
+	settingsSvc         model.SettingsService
+	projectRepo         model.ProjectRepository
+	envRepo             model.EnvironmentRepository
+	invitationRepo      model.ProjectInvitationRepository
+	githubClientFactory model.GithubClientFactory
```
jan-zabloudil, line 26: A GitHub client is created in the service only if all conditions are met (integration enabled, project has repository...) Therefore, the service accepts a `GitHubClientFactory` interface.

github/github.go
```diff
@@ -0,0 +1,41 @@
...
+		Client: github.NewClient(nil).WithAuthToken(tkn),
+	}
+}
+
+func (c *Client) ListTagsForRepository(ctx context.Context, repo svcmodel.GithubRepository, limit int) ([]svcmodel.GitTag, error) {
+	t, _, err := c.Client.Repositories.ListTags(
+		ctx,
+		repo.OwnerSlug,
+		repo.RepositorySlug,
+		&github.ListOptions{PerPage: limit},
```
bolshakoff, line 34: Don't we also want to specify the page?

Or use a much bigger limit and document it, for example right here, above the method.
(Obviously, would be considered a short-run solution, and "add pagination" would be a separate task, but that's fine.)

And since the `githubRepositoryTagsLimit` is used only in one place, maybe even simply move it here, to the client...

service/project.go
bolshakoff: That's a valid point; I think it deserves a comment.

service/project.go
```diff
@@ -2,32 +2,45 @@ package service
...
 }
 
 func NewProjectService(
 	authSvc model.AuthService,
+	settingsSvc model.SettingsService,
 	projectRepo model.ProjectRepository,
 	envRepo model.EnvironmentRepository,
 	invitationRepo model.ProjectInvitationRepository,
+	factory model.GithubClientFactory,
```
bolshakoff, line 35: I'd rename it to `githubFactory`.

github/mock/mock.go
```diff
@@ -0,0 +1,31 @@
+package mock
```
bolshakoff, line 1: Consider using the [mock that is recommended by go-github](https://github.com/google/go-github?tab=readme-ov-file#testing-code-that-uses-go-github).

service/project.go
bolshakoff: Or consider simply having `githubClient` as a pointer.

bolshakoff: (to get rid of the factory which feels overly complex)

jan-zabloudil: Thanks, it was actually very easy to implement it without the factory.

pkg/apierrors/apierrors.go
```diff
@@ -151,14 +192,21 @@ func IsUnprocessableModelError(err error) bool {
...
+	return IsErrorWithCode(err, errCodeForbiddenInsufficientUserRole) ||
+		IsErrorWithCode(err, errCodeGithubIntegrationForbidden)
 }
 
 func IsConflictError(err error) bool {
 	return IsErrorWithCode(err, errCodeEnvironmentDuplicateName) ||
 		IsErrorWithCode(err, errCodeProjectInvitationAlreadyExists)
 }
+
+func IsPreconditionFailedError(err error) bool {
```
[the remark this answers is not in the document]
jan-zabloudil: Thanks! These errors is something Marek recommended but I think you also made some pretty valid comments. Let's discuss it here - https://github.com/jan-zabloudil/release-manager/pull/29#discussion_r1583100401
(I'll try to write a summary similar to the one we used when discussing the service structure)).

pkg/githuberrors/githuberrors.go
```diff
@@ -0,0 +1,74 @@
...
+func IsErrorWithCode(err error, code string) bool {
+	var githubErr *GithubError
+	if errors.As(err, &githubErr) {
+		return githubErr.Code == code
+	}
+
+	return false
+}
+
+func IsUnauthorizedError(err error) bool {
```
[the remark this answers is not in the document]
jan-zabloudil: I will address it in https://github.com/jan-zabloudil/release-manager/issues/25

transport/util/errormapper.go
```diff
@@ -17,6 +17,8 @@ func ToResponseError(err error) *responseerrors.ResponseError {
 		return responseerrors.NewUnprocessableEntityError().Wrap(err)
 	case apierrors.IsConflictError(err):
 		return responseerrors.NewConflictError().Wrap(err)
+	case apierrors.IsPreconditionFailedError(err):
```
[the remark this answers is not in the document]
jan-zabloudil: > High-level comment - as indicated earlier, consider gathering transport layer error mapper and transport layer errors into one file / one package.
Ideally under transport layer.

Good point, will be addressed in https://github.com/jan-zabloudil/release-manager/issues/25

> Second comment - is it really a Precondition Failed error? From a quick read, this is a completely different kind of error... 🤔

Yes, you are right. If GitHub is not enabled or the GitHub repository is not configured, a 404 error is returned now (the resource is not found because it is not configured). Feel free to suggest a different response code if you think there is a better choice.

github/github.go
jan-zabloudil: In the end, I removed the list options configuration and kept the default settings. Now, 30 tags will be returned by default.

I think it's okay for the MVP. The last 30 tags should be more than enough since you'll usually choose the newest ones when linking the release to the source code. Additionally, if the client wants to display fewer tags, this can be handled on the client side.

github/mock/mock.go
jan-zabloudil: I created an issue for myself and I would explore this option later on and consider it, thanks!
https://github.com/jan-zabloudil/release-manager/issues/31

github/github.go
```diff
@@ -24,12 +24,14 @@ func (c *Client) SetToken(token string) {
 	c.client = c.client.WithAuthToken(token)
 }
 
-func (c *Client) ListTagsForRepository(ctx context.Context, repo svcmodel.GithubRepository, limit int) ([]svcmodel.GitTag, error) {
+func (c *Client) ListTagsForRepository(ctx context.Context, repo svcmodel.GithubRepository) ([]svcmodel.GitTag, error) {
+	// Last 30 tags are fetched when no options are provided
```
bolshakoff: Love the comment, explanation and link to their docs!

But consider simply putting max value here, i.e. `100`.

bolshakoff (approved):

PR #30, opened 2024-04-30
feat: send project invitation email using resend

jan-zabloudil:
Send invitation email using [Resend](https://resend.com)


transport/model/interface.go
```diff
@@ -24,7 +25,7 @@ type ProjectService interface {
 }
 
 type ProjectMembershipService interface {
-	CreateInvitation(ctx context.Context, c svcmodel.CreateProjectInvitationInput, authUserID uuid.UUID) (svcmodel.ProjectInvitation, error)
+	CreateInvitation(ctx context.Context, m *background.Manager, c svcmodel.CreateProjectInvitationInput, authUserID uuid.UUID) (svcmodel.ProjectInvitation, error)
```
jan-zabloudil, line 28: I prefer to pass the manager to the service explicitly rather than including it as part of the service struct. However, I would be happy to hear your opinion on this.

resend/resend.go
```diff
@@ -0,0 +1,49 @@
...
+
+func NewClient(isLive config.IsLive, cfg config.ResendConfig) *Resend {
+	return &Resend{
+		client:     resend.NewClient(cfg.APIKey),
+		reqBuilder: NewEmailRequestBuilder(isLive, cfg),
+	}
+}
+
+func (r *Resend) SendProjectInvitationEmail(ctx context.Context, input svcmodel.ProjectInvitationEmailInput) error {
+	// TODO implement proper email template (both html and text) with proper message and magic link, this is just a starting point!
```
jan-zabloudil, line 27: To keep PR smaller, I will add proper email template in separate PR.

service/project_membership.go
```diff
@@ -9,36 +9,43 @@ import (
...
 	"github.com/google/uuid"
+	"go.strv.io/background"
+	"go.strv.io/background/task"
 )
 
 type ProjectMembershipService struct {
 	authSvc    model.AuthService
 	projectSvc model.ProjectService
 	repository model.ProjectInvitationRepository
+	mailer     model.Mailer
```
bolshakoff, line 20: Here and for the rest of the services above, it feels weird that interfaces are a part of model package.

Wouldn't it be smoother to at least have it simply near `service.go`?.. 🤔

transport/model/interface.go
bolshakoff: Strictly speaking, here we are stating an interface from the transport layer perspective. The fact that invitation creation involves sending an email AND sending it asynchronously shouldn't be transport layer's business at all.

Moreover, should the background manager be a part of service? Probably no; in principle, it only states an interface that there must be somebody who will send an email invitation.
Perhaps here, the most accurate would be to state something like:
```go
type EmailSender interface {
    SendEmailAsync(...)
}
```
So that the service will know that this call will run asynchronously and thus won't take too much time.
Ideally without bothering that it will need a background manager - this will be solved by the `EmailSender` (or in our case `Mailer`) implementation itself. Because it doesn't have to be specifically background manager, right?

bolshakoff: And so, I'd probably not inject the `TaskManager` in `main`, but rather have it directly as a part of `Resend` struct.

bolshakoff: (depending on the expected usage etc.; maybe soon it will be used somewhere else and the intended usage is to initialize `TaskManager` in `main` - in this case still, I'd just feed it directly to the `Resend`...)

config/config.go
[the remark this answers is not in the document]
bolshakoff: OK

bolshakoff (commented):
In general, might be useful or even necessary to store the info about sent emails and their status into db.

But here, definitely no need, it would be an overkill.

jan-zabloudil:
> In general, might be useful or even necessary to store the info about sent emails and their status into db.
> 
> But here, definitely no need, it would be an overkill.

Good point, but agree it is not needed in this case. And Resend actually shows you status of all sent emails (including the failed ones), so should be enough for MVP :)

---

Sorry, I tried to delete duplicated comments but GitHub actually deleted all comments 🙈

service/project_membership.go
jan-zabloudil: I created a separate PR #32

resend/resend.go
```diff
@@ -0,0 +1,52 @@
...
+func NewClient(manager *background.Manager, isLive config.IsLive, cfg config.ResendConfig) *Resend {
+	return &Resend{
+		taskManager: manager,
+		client:      resend.NewClient(cfg.APIKey),
+		reqBuilder:  NewEmailRequestBuilder(isLive, cfg),
+	}
+}
+
+func (r *Resend) SendEmailAsync(ctx context.Context, subject, text, html string, recipients ...string) {
+	req := r.reqBuilder.
```
jan-zabloudil, line 29: Resend pkg now only cares about sending an email.

service/model/interface.go
```diff
@@ -58,3 +58,11 @@ type ProjectInvitationRepository interface {
 type ProjectService interface {
 	Get(ctx context.Context, projectID, authUserID uuid.UUID) (Project, error)
 }
+
+type ProjectInvitationSender interface {
+	SendProjectInvitation(ctx context.Context, input ProjectInvitationInput)
+}
```
jan-zabloudil, line 64: Interface for sending project invitations. The service now does not need to know whether the invitation is sent via email or another channel

service/email.go
jan-zabloudil, line 1: Email service implements `ProjectInvitationSender`.

I have realized (thank for the comments) that the resend package should focus only on sending the email, without concerning itself with the email's appearance or content. Therefore, I have added an `EmailService` to handle these aspects.

bolshakoff (approved):

config/config.go
```diff
@@ -9,26 +9,36 @@ import (
...
 
-type supabaseConfig struct {
+type SupabaseConfig struct {
 	APIURL    string `env:"API_URL, required"`
 	SecretKey string `env:"SECRET_KEY, required"`
 }
 
-type serviceConfig struct {
+type ServiceConfig struct {
+	IsLive   IsLive         `env:"IS_LIVE, default=false"`
```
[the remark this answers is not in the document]
jan-zabloudil: Yes, I was struggeling with this. 

I was also considering adding some field `SendToRealRecipients` directly to `ResendConfig`. 

```
type ResendConfig struct {
	APIKey               string `env:"API_KEY, required"`
	Sender               string `env:"SENDER, required"`
	TestRecipient        string `env:"TEST_RECIPIENT, required"`
	SendToRealRecipients bool   `env:"SEND_TO_REAL_RECIPIENTS, default=false"`
}
```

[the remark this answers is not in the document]
bolshakoff: Looks better, but maybe the default value should be `true`; 1) it's 99% in the long run; 2) could be an unnecessary source of bug in the future (if we forget to set it to `true` on production, for example)...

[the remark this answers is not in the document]
jan-zabloudil: To be honest, I would be more afraid of the opposite case when you accidentally send a real email during development

I would meet in the middle :D and remove the default option. This way, you would have to set it explicitly in the config.

PR #32, opened 2024-05-01
refactor: improve transport layer structure & move interfaces to consumer packages

jan-zabloudil:
I want to move interfaces to consumer (where interfaces are actually use) packages. 

**Transport**

In the transport layer, I first reorganize the code into separate packages

```
└── transport/
    ├── handler
    ├── middleware
    ├── model
    └── util
```

And handler and middleware define their interfaces.

----

In service layer, interfaces are now defined in consumer package as well.


service/service.go
```diff
@@ -1,6 +1,64 @@
...
+}
+
+type authGuard interface {
+	AuthorizeAdminRole(ctx context.Context, userID uuid.UUID) error
+	AuthorizeRole(ctx context.Context, userID uuid.UUID, role model.UserRole) error
+}
+
+type projectGetter interface {
+	Get(ctx context.Context, projectID, authUserID uuid.UUID) (model.Project, error)
+}
```
jan-zabloudil, line 61: This is just temporary because the `Project` and `ProjectInvitation` services will be merged into one, and this interface will no longer be necessary.

I borrowed name `projectGetter` from [this book](https://www.amazon.com/100-Mistakes-How-Avoid-Them/dp/1617299596).

I initially felt it was overly descriptive. However, It improved code readability: (and we were actually discussing exactly this call)

```
if _, err := s.projectGetter.Get(....); err != nil {

}
```

bolshakoff: >This is just temporary because the Project and ProjectInvitation services will be merged into one, and this interface will no longer be necessary.

💯 

>I borrowed name projectGetter from [this book](https://www.amazon.com/100-Mistakes-How-Avoid-Them/dp/1617299596).

Yeah, a must-read book that I.. haven't read yet! 😁 
The more interesting it will be during reviews - as always, feel free to counter-argue me. 😈 

This one is one of my favorite topics, actually. 🔥 
Academically, I think you are absolutely correct.
1) You are using interfaces
2) You are keeping them as slim and explicit as possible
So, the resulting potential for decoupling etc. is maximum or almost maximum.

I once tried to promote a similar thing at my project as well, but the counter-argument was:
- Yes, but then it's too scattered, too complex. Unnecessarily scattered and complex, because we know FOR SURE that it's going to be (in this case) `ProjectService` who will implement `Get()` method, `Delete()` method etc. In other words, we are practically guaranteed to not have a `ProjectGetterService` and `ProjectDeleterService`.

If we take this argument to extreme, we might say - well, if we are guaranteed to have one and only one implementation, one `ProjectService`, why do we need an interface here at all?
If you are really into that, I won't even spoil the "answers". 😁 
But one of them can be nicely and clearly inferred from the classic [Uncle Bob's talk about SOLID principles](https://www.youtube.com/watch?v=zHiWqnTWsn4). Must watch.

bolshakoff (approved):

service/service.go
jan-zabloudil: > Yeah, a must-read book that I.. haven't read yet! 😁

I havent read the whole book actually yet. Some chapters I have found too heavy for me yet as I am new to Go :D. However, I did enjoy the chapters focused on Code and Project Organization.

And I actually agree with you. I think the fact I need interface `projectGetter` more indicates that it would be better to structure service differently. However, in some cases, having a slim interface makes sense to me, such as with `projectInvitationSender` in #30.

PR #34, opened 2024-05-01
feat: create project member upon user registration

jan-zabloudil:
I don't control user sign-ups via Google SSO; therefore, the only way to create a project member upon user registration is via a PostgreSQL function.

When a user accepts an invitation as an already registered user, I will handle the creation of the project member in Go.


supabase/migrations/20240501085903_handle_new_user_update.sql
```diff
@@ -0,0 +1,30 @@
...
+    VALUES (
+               new.id,
+               new.email,
+               new.raw_user_meta_data ->> 'name',
+               new.raw_user_meta_data ->> 'picture',
+               new.created_at,
+               new.updated_at
+           );
+
+    -- If user has accepted an invitation(s) before signing up, add them to the project(s)
```
bolshakoff, line 19: Is it possible to add this block below as a separate function?

So that the migration file basically contains only changes.

bolshakoff (approved):

supabase/migrations/20240501085903_handle_new_user_update.sql
jan-zabloudil: I could create two standalone functions, then wrap them in another function that would be triggered.

```
CREATE OR REPLACE FUNCTION public.xxx()
RETURNS TRIGGER AS $$
BEGIN
    PERFORM public.handle_new_user();

    PERFORM public.create_project_member();
END;
$$ LANGUAGE plpgsql;
```

```
CREATE TRIGGER on_auth_user_created
AFTER INSERT ON auth.users
FOR EACH ROW EXECUTE PROCEDURE public.xxx()
```

But I will stick with my current solution.

PR #37, opened 2024-05-04
refactor: simplify github logic

jan-zabloudil:
While implementing a new functions that access the GitHub API, I realized that `ProjectService/ProjectModel` was handling too much GitHub integration logic, such as parsing GitHub URLs, checking settings, and handling errors. 

To streamline this, I moved all GitHub-specific logic to `GithubService`. Now, ProjectService can simply call:

```
ListTags(ctx context.Context, repoURL url.URL) ([]model.GitTag, error)
```

With this change, `ProjectService/Model` is only responsible for providing the project's github repository URL (even without performing any extra URL validation,  `GithubService` will handle these aspects). 

This refactoring will make it easier to add new GitHub integration functionality.

-----

If you agree with this approach I will also add tests.


jan-zabloudil:
I have realized that since the GitHub client is used for reading and writing data, it can simply be move into the repository.

```
└── repository/
    ├── db/
    │   └── ...
    ├── githubapi/
    │   ├── model
    │   ├── util
    │   ├── git.go
    │   └── release.go
    └── repository.go
```

Currently, there is only a `GitRepository` for accessing all Git objects, such as tags and commits. Then `ReleaseRepository` can be added for creating github release, generating release notes etc.

repository/githubapi/git.go
```diff
@@ -0,0 +1,41 @@
...
+	// If the number of pages is not specified in the list options, only one page will be fetched
+	// https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-tags
+	t, _, err := client.Repositories.ListTags(
+		ctx,
+		repo.OwnerSlug,
+		repo.RepositorySlug,
+		&github.ListOptions{PerPage: tagsToFetch},
+	)
+	if err != nil {
+		return nil, util.ToResponseError(err)
```
jan-zabloudil, line 37: And as you mentioned @bolshakoff, here simply makes sense to return `responseerror`.

bolshakoff: I see. Yeah, I have "to an api layer", which is not precise, because I meant service layer.
Sorry for a confusion.

So, in general, the principle is:
- domain / svc knows only about itself and its errors
- repo translates to domain errors
- transport translates from domain errors to transport errors

bolshakoff: What you had before, i.e. translation from `GithubError` to `APIERror` in svc wasn't too bad.
But as you can see, it becomes too many imports, too many external knowledge for the domain. But domain should ideally know only about itself, so that you always have a guarantee that HERE you are dealing purely with the business logic.
So, for now, I'd move it back.

bolshakoff: >But as you can see, it becomes too many imports, too many external knowledge for the domain.

So, as you can see, the approach where it's up to other layers to translate to/from domain/svc layer might be pleasant.
As a result, there is a simple "sun" topology, where everybody imports domain models and domain errors.. everybody serves to domain.. and that's fine, because for all repositories, 3rd party services and transport layer it's just one import, as opposed to the situation when domain/service layer might have to import tons of different modules and deal with tons of different types of errors...

bolshakoff:
Regarding the change here, I'd say the biggest problem now is that `github` being in service / domain layer suggests that GitHub is a domain in the application. But it isn't.
It's basically a repository of git tags, nothing else.

So, while making it a repository is a valid move, if it's purely a repository, it should be sufficient to have it as a repository, and nothing else.

Any non-trivial logic like settings check belong to the `project` domain, so I'd not overthink it.

Also, on the mechanical level, it is now a bit too complicated.
A very simple thing like listing github tags requires so many abstractions. It might sound correct in theory, but in practice, the code becomes less smooth, less sharp.

Also, the question is, will we need `github.Client` for other, non-trivial things?
Then perhaps it would be better to have as we did - simply as a 3rd party dependency.

bolshakoff (changes requested):
As I say, now it's pushed a bit too far.

jan-zabloudil:
Thanks @bolshakoff , I get your point!

> Also, the question is, will we need github.Client for other, non-trivial things?

Yes, we will need it for other operations, such as creating a GitHub release and auto-generating release notes etc.

----

## I suggest the following modifications 

First, `settingsGetter` could have `GetGithubToken()` function (instead of `GetGithubSettings`). 
- This way you can simplify logic in the project service (and later on in release service). 
- Additionally, it makes sense for the setting service to be responsible for providing a token or returning a corresponding error if the settings is not enabled.

**Before**

```
github, err := s.settingsGetter.GetGithubSettings(ctx)
if err != nil {
     return nil, err
}

if !github.Enabled {
     return nil, apierrors.NewGithubIntegrationNotEnabledError()
}
```

**Now**
```
tkn, err := s.settingsGetter.GetGithubToken(ctx)
if err != nil {
...
}
```

### Repository idea
 
I would drop the repository idea and simply have `pkg/github`. Service would define interface:

```
type githubManager struct {
    ReadTagsFromReposirory(ctx context.Context, token string, repoURL url.URL) ([]model.GitTag, error)
}
```

- This interface will be implemented by `pkg/github`. 
- It would be `pkg/github` responsibility to check and parse repoURL.
- This way `/service/model/github.go` and `/service/github.go` can be removed.
- And package would return specific errors such as:
   - access denied (invalid token or token without permission)
   - invalid github repo url 
   - resource not found
   - ....

And service function looks quite simply now:

```
func (s *ProjectService) ListGithubRepositoryTags(...) ([]model.GitTag, error) {
	// TODO add project member authorization

	project, err := s.GetProject(ctx, projectID, authUserID)
	if err != nil {
		return nil, err
	}

	if !project.IsGithubConfigured() {
		return nil, apierrors.NewGithubRepositoryNotConfiguredForProjectError()
	}

        tkn, err := s.settingsGetter.GetGithubToken(ctx)
        if err != nil {
           ...
        }

	t, err := s.githubManager.ReadTagsFromRepository(ctx, tkn, project.GithubRepositoryURL)
        if err != nil {
          ...
        }

        return t, nil
}
```

bolshakoff:
It's a good direction, and looks correct on the paper, but I think at least on the GitHub side it deserves an even more simple approach:
- Instead of github manager and github client, I'd go.. or at least start (since we now have basically only one trivial operation) with only github client. And that's OK that project service will refer almost directly to github client, because 1) it's simple, 2) it's sharp
- Not sure it's a good idea to have github client in `pkg` though. It feels more natural to have it on the top-level, similar to `resend`. Another argument - similarly to `resend`, github client also deserves a mock implementation.

github/client.go
```diff
@@ -0,0 +1,62 @@
...
+		ctx,
+		repo.OwnerSlug,
+		repo.RepositorySlug,
+		&github.ListOptions{PerPage: tagsToFetch},
+	)
+	if err != nil {
+		var githubErr *github.ErrorResponse
+		if errors.As(err, &githubErr) {
+			switch githubErr.Response.StatusCode {
+			case http.StatusUnauthorized:
```
jan-zabloudil, line 49: I am trying the approach where the GitHub package directly returns API errors

I want to explore this approach more and the suggest proper refactor here -> #25

bolshakoff (approved):
Niice! Love the simplicity. 🤩

PR #38, opened 2024-05-04
feat: add endpoints for listing project members and deleting them

jan-zabloudil:
To keep PR smaller, other endpoints will be added in separate PRs.

### DIFF
--- a/service/project.go
+++ b/service/project.go
@@ -370,6 +370,48 @@ func (s *ProjectService) RejectInvitation(ctx context.Context, tkn cryptox.Token
 	return s.repo.DeleteInvitation(ctx, invitation.ID)
 }
 
+func (s *ProjectService) ListMembers(ctx context.Context, projectID, authUserID uuid.UUID) ([]model.ProjectMember, error) {
+	if err := s.authGuard.AuthorizeAdminRole(ctx, authUserID); err != nil {
+		return nil, err
+	}
+
+	exists, err := s.projectExists(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, apierrors.NewProjectNotFoundError()
+	}
+
+	return s.repo.ReadMembersForProject(ctx, projectID)
+}
+
+func (s *ProjectService) DeleteMember(ctx context.Context, projectID, userID, authUserID uuid.UUID) error {
+	if err := s.authGuard.AuthorizeAdminRole(ctx, authUserID); err != nil {
+		return err
+	}
+
+	exists, err := s.projectExists(ctx, projectID)
+	if err != nil {
+		return err
+	}
+	if !exists {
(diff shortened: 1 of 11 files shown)


service/project.go
```diff
@@ -370,6 +370,48 @@ func (s *ProjectService) RejectInvitation(ctx context.Context, tkn cryptox.Token
...
+	if err != nil {
+		switch {
+		case dberrors.IsNotFoundError(err):
+			return apierrors.NewProjectMemberNotFoundError().Wrap(err)
+		default:
+			return err
+		}
+	}
+
+	return s.repo.DeleteMember(ctx, projectID, userID)
```
bolshakoff, line 412: The user can get deleted between these two calls.

So, while I really like that it is plain and obvious, and quite smooth, here we need to add some sharpness - basically fix the race condition above.
If the user that is being deleted does not exist, there should be a corresponding pg error.

service/project.go
```diff
@@ -370,6 +370,48 @@ func (s *ProjectService) RejectInvitation(ctx context.Context, tkn cryptox.Token
...
+		return err
+	}
+
+	exists, err := s.projectExists(ctx, projectID)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return apierrors.NewProjectNotFoundError()
+	}
```
bolshakoff, line 400: Actually, a similar thing here.
Meanwhile, the project might get deleted, soo this check is not sharp enough.

In the case of member deletion, I think everything can be simplified to one call and returning a `ProjectMemberNotFoundError`.

But when it comes to listing.. see next comment...

service/project.go
```diff
@@ -370,6 +370,48 @@ func (s *ProjectService) RejectInvitation(ctx context.Context, tkn cryptox.Token
...
+
+	exists, err := s.projectExists(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, apierrors.NewProjectNotFoundError()
+	}
+
+	return s.repo.ReadMembersForProject(ctx, projectID)
```
bolshakoff, line 386: When it comes to listing, perhaps a more correct implementation here is to query the members, and if it is empty check whether it is because the project does not exist (`404`) or simply there are indeed no members in the project (`200`).

bolshakoff: But it's a minor thing; and if it is in more places, I'd maybe leave it to a separate PR or maybe issue, because the opposite argument here is also that it's such a rare and non-critical case, that not worth it...

bolshakoff (approved):

jan-zabloudil:
Thanks, good point! I’ve created issue #47, where I suggested a method to prevent the race condition. Please take a look.

I will implement the fix from #47 once all open PRs are merged. If you notice this race condition occurring with other PRs as well, there's no need to mention it, as it will be addressed.

PR #42, opened 2024-05-05
feat: check if user is already project member when creating invitation

jan-zabloudil:


repository/project.go
```diff
@@ -351,3 +351,25 @@ func (r *ProjectRepository) DeleteMember(ctx context.Context, projectID, userID
...
 	return nil
 }
+
+func (r *ProjectRepository) ReadMemberByEmail(ctx context.Context, projectID uuid.UUID, email string) (svcmodel.ProjectMember, error) {
+	// When querying data via Supabase Rest API (that uses PostgREST), only left joins are possible
+	// But in this case inner join is needed -> join project_members with users and return only records where user email matches
+	// 3 possible solutions:
+	// 1. Write postgres function that will return joined data (creates extra overhead)
+	// 2. Use left join and then loop over result and filter out records where user email does not match/user is empty (not efficient, inelegant)
+	// 3. In first query get user by email and then query project_members by project_id and user_id
```
bolshakoff, line 361: Isn't it in another PR where there is a `ReadUserByEmail()` method?
I'd prefer to simply do two calls from svc, as opposed to the current, "nested" implementation.

bolshakoff (approved):

PR #43, opened 2024-05-05
feat: add endpoint for updating project member role 

jan-zabloudil:
Other endpoints are implemented in #38

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -484,26 +484,7 @@ paths:
               $ref: '#/components/responses/UnauthorizedErrorResponse'
         '404':
               $ref: '#/components/responses/NotFoundErrorResponse'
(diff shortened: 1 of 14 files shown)


api-doc.yaml
```diff
@@ -484,34 +484,16 @@ paths:
               $ref: '#/components/responses/UnauthorizedErrorResponse'
         '404':
               $ref: '#/components/responses/NotFoundErrorResponse'
-  /members/{member-id}:
```
jan-zabloudil, line 487: There is no need for an endpoint to fetch a specific member since only the project member's role can be updated, or the member can be deleted. On the frontend, a list of all members will be displayed, and users can directly update the role or delete members from this list

bolshakoff (approved):

PR #44, opened 2024-05-05
feat: add create release endpoint

jan-zabloudil:
To keep the PR smaller, this PR introduces a very basic Release entity, consisting only of a name and release notes, and adds an endpoint for creating a release. 

Once this is merged, more features can be developed simultaneously, such as new endpoints, Slack notifications, and linking source code (GitHub integration).

### DIFF
--- a/supabase/migrations/20240505115913_releases.sql
+++ b/supabase/migrations/20240505115913_releases.sql
@@ -1,0 +1,13 @@
+CREATE TABLE public.releases (
+     id UUID NOT NULL PRIMARY KEY,
+     project_id UUID NOT NULL REFERENCES public.projects ON DELETE CASCADE,
+     release_title TEXT NOT NULL,
+     release_notes TEXT,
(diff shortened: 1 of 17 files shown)


supabase/migrations/20240505115913_releases.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE public.releases (
+     id UUID NOT NULL PRIMARY KEY,
+     project_id UUID NOT NULL REFERENCES public.projects ON DELETE CASCADE,
+     name TEXT NOT NULL,
```
jan-zabloudil, line 4: Would you make release name unique per project?

bolshakoff: I think yes; in case we decide to allow duplicate names, we can always drop the constraint, but in the opposite direction would be much more difficult.
Also, I'd probably name it `release_title`, if it better reflects the role of this field.

PR #45, opened 2024-05-05
refactor: move resend client to pkg

jan-zabloudil:


pkg/resend/client.go
```diff
@@ -41,7 +41,7 @@ func (c *Client) SendEmailAsync(ctx context.Context, subject, text, html string,
 		Fn: func(ctx context.Context) error {
 			_, err := c.client.Emails.SendWithContext(ctx, req)
 			if err != nil {
-				return emailerrors.NewEmailSendingFailedError().Wrap(err)
+				return fmt.Errorf("failed to send email via Resend: %w", err)
```
jan-zabloudil, line 44: Emails are sent asynchronously, and potential errors are only logged, not handled in the service. Therefore, there is no need to have `pkg/emailerrors`.

jan-zabloudil:
Not a good idea to move resend to `pkg` after all.

PR #46, opened 2024-05-06
feat: add slack integration

jan-zabloudil:
This PR introduces the ability to send Slack messages from the app in the following format: message + [message attachment](https://api.slack.com/reference/messaging/attachments)

Snímek obrazovky 2024-05-06 v 9 51 58

### DIFF
--- a/service/model/project.go
+++ b/service/model/project.go
@@ -38,21 +38,21 @@ type UpdateProjectInput struct {
 }
 
 type ReleaseNotificationConfig struct {
-	Message         string
-	ShowProjectName bool
-	ShowReleaseName bool
-	ShowChangelog   bool
-	ShowDeployments bool
-	ShowSourceCode  bool
+	Message          string
+	ShowProjectName  bool
+	ShowReleaseTitle bool
+	ShowReleaseNotes bool
+	ShowDeployments  bool
+	ShowSourceCode   bool
 }
 
 type UpdateReleaseNotificationConfigInput struct {
-	Message         *string
-	ShowProjectName *bool
-	ShowReleaseName *bool
-	ShowChangelog   *bool
-	ShowDeployments *bool
-	ShowSourceCode  *bool
+	Message          *string
+	ShowProjectName  *bool
+	ShowReleaseTitle *bool
+	ShowReleaseNotes *bool
+	ShowDeployments  *bool
+	ShowSourceCode   *bool
 }
 
 func NewProject(c CreateProjectInput) (Project, error) {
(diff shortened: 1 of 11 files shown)


service/service.go
```diff
@@ -54,6 +54,7 @@ type authGuard interface {
 
 type settingsGetter interface {
 	GetGithubSettings(ctx context.Context) (model.GithubSettings, error)
+	GetSlackSettings(ctx context.Context) (settings model.SlackSettings, err error)
```
jan-zabloudil, line 57: As suggested in #37, it might be better to have `GetSlackToken(ctx context.Context) (string, error)`.

pkg/slack/client.go
```diff
@@ -0,0 +1,49 @@
+package slack
```
bolshakoff, line 1: I presume similarly to `resend` this will be moved to a good old top-level package.

bolshakoff (changes requested):
Again, similar to GitHub, now it seems overcomplicated.

I'd find it much more smooth and sharp if you simply had a top-level `slack` package with `client` and `mock`, and simply have `SlackClient` whenever it is needed.

Where to have the msg model? If it doesn't have anything Slack-specific, then in domain, otherwise in Slack.. But overall, doesn't seem to be a big deal...

jan-zabloudil:
To keep PR small, only `SlackClient` is added. Later on, client will be used in `ReleaseService`. Release service will define interface 

```
type slackClient interface {
	SendReleaseNotification(ctx context.Context, token, channelID string, n model.ReleaseNotification)
}
```

service/model/release.go
```diff
@@ -53,3 +53,28 @@ func (r *Release) Validate() error {
 
 	return nil
 }
+
+type ReleaseNotification struct {
+	Message      string
+	ProjectName  *string
+	ReleaseTitle *string
+	ReleaseNotes *string
+}
```
jan-zabloudil, line 62: More fields will be added once the whole release functionality is implemented.

service/model/project.go
```diff
@@ -50,21 +50,21 @@ type UpdateProjectInput struct {
...
-	Message         string
-	ShowProjectName bool
-	ShowReleaseName bool
-	ShowChangelog   bool
-	ShowDeployments bool
-	ShowSourceCode  bool
+	Message          string
+	ShowProjectName  bool
+	ShowReleaseTitle bool
+	ShowReleaseNotes bool
```
jan-zabloudil, line 56: Fields in `ReleaseNotificationConfig` were renamed to match fields in Release model.

bolshakoff (approved):

PR #49, opened 2024-05-09
refactor: improve naming in resend package; drop emailerrors

jan-zabloudil:
@bolshakoff  Following the approach suggested for `Slack` and `Github`, I should also remove the `EmailService` and directly use the email client (using interface of course) within the` ProjectService`. `ProjectService` would prepare the email content and then call 

`emailClient.SendEmailAsync(...)`

Do you agree?


resend/client.go
```diff
@@ -39,14 +39,14 @@ func (r *Resend) SendEmailAsync(ctx context.Context, subject, text, html string,
 			"task": "sending email",
 		},
 		Fn: func(ctx context.Context) error {
-			_, err := r.client.Emails.SendWithContext(ctx, req)
+			_, err := c.client.Emails.SendWithContext(ctx, req)
 			if err != nil {
-				return emailerrors.NewEmailSendingFailedError().Wrap(err)
+				return fmt.Errorf("failed to send email via Resend: %w", err)
```
jan-zabloudil, line 44: Emails are sent asynchronously, and potential errors are only logged, not handled in the service. Therefore, there is no need to have `pkg/emailerrors`.

bolshakoff (approved):
Nice.

Issue #47, opened 2024-05-07
Prevent raise condition in project service 


jan-zabloudil:
Applies to:
- project invitations 
- project environments
- project members

New approach not only prevent raise condition but also:
- reduce db calls
- simplifies the code 

## Listing resources (`/projects/{project_id}/invitations`)

**Current implementation**

```

func (s *ProjectService) ListInvitations(...) (...) {
	...

	exists, err := s.projectExists(ctx, projectID)
	if err != nil {
		return nil, err
	}
	if !exists {
		return nil, apierrors.NewProjectNotFoundError()
	}

        // between these two calls (s.projectExists calls DB), project can be deleted

	return s.repo.ReadAllInvitationsForProject(ctx, projectID)
}

```

**Desired**

```
func (s *ProjectService) ListInvitations(...) ([]model.ProjectInvitation, error) {
	...
        
        i, err :=  s.repo.ReadAllInvitationsForProject(ctx, projectID)
        if err != nil {
		return nil, err
	}

        if len(i) == 0 {
               // check if project exists and return err if not
        }

	return i, nill
}

```

## Get specific resource (`/projects/{id}/environments/{environment_id}`)

**Current implementation**

```
func (s *ProjectService) GetEnvironment(...) (model.Environment, error) {
      .....

	_, err := s.GetProject(ctx, projectID, authUserID)
	if err != nil {
		return model.Environment{}, err
	}

        // Project can be deleted between those two calls anyway

	env, err := s.repo.ReadEnvironment(ctx, envID)
	if err != nil {
		switch {
		case dberrors.IsNotFoundError(err):
			return model.Environment{}, apierrors.NewEnvironmentNotFoundError().Wrap(err)
		default:
			return model.Environment{}, err
		}
	}

	return env, nil
}
```

**Desired**

- Just one db call
- Ensure the database call includes the `project_id`.
- The API does not specifically verify the existence of the project; instead, it returns `EnvironmentNotFoundError()` in all cases 
- However, I believe this is fine and still enough user-friendly API

```
func (s *ProjectService) GetEnvironment(...) (model.Environment, error) {
      .....

	env, err := s.repo.ReadEnvironmentForProject(ctx, projectID, envID)
	if err != nil {
		switch {
		case dberrors.IsNotFoundError(err):
			return model.Environment{}, apierrors.NewEnvironmentNotFoundError().Wrap(err)
		default:
			return model.Environment{}, err
		}
	}

	return env, nil
}
```

[...]

PR #51, opened 2024-05-09
feat: add db pool and use it in release repository

jan-zabloudil:
Release repository implementation uses db pool instead of supabase client.

Readme will be updated in separate PR -> #50


repository/release.go
```diff
@@ -2,36 +2,56 @@ package repository
...
+		"createdAt":    rls.CreatedAt,
+		"updatedAt":    rls.UpdatedAt,
+	}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReleaseRepository) Read(ctx context.Context, id uuid.UUID) (svcmodel.Release, error) {
```
jan-zabloudil, line 44: Added a `Read` function to demonstrate how data will be retrieved from the database. This function is not currently used in the service  at the moment (but it will be of course).

I have also checked [the article](https://www.strv.com/blog/creating-a-project-template-go-with-database-storage). I will start with very simple implementation and helper functions will be added gradually.

repository/release.go
```diff
@@ -2,36 +2,57 @@ package repository
...
 	return nil
 }
+
+func (r *ReleaseRepository) Read(ctx context.Context, id uuid.UUID) (svcmodel.Release, error) {
+	var rls model.Release
+
+	err := pgxscan.Get(ctx, r.dbpool, &rls, query.ReadRelease, pgx.NamedArgs{"id": id})
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return svcmodel.Release{}, apierrors.NewReleaseNotFoundError().Wrap(err)
```
jan-zabloudil, line 51: Repository returns apierror directly

jan-zabloudil: The new approach is that all packages that "serve" the service will now return a service error. 

Currently, the errors are located in `pkg/apierrors`, but once the refactoring is complete, these errors will be moved to the `service` package, and `responseerrors` will be relocated to the transport package.

repository/release.go
```diff
@@ -2,36 +2,57 @@ package repository
...
+	})
 	if err != nil {
-		return util.ToDBError(err)
+		return err
 	}
 
 	return nil
 }
+
+func (r *ReleaseRepository) ReadForProject(ctx context.Context, projectID, releaseID uuid.UUID) (svcmodel.Release, error) {
```
jan-zabloudil, line 45: Added a Read function to demonstrate how data will be retrieved from the database. This function is not used in the service at the moment (but it will be of course).

I have also checked [the article](https://www.strv.com/blog/creating-a-project-template-go-with-database-storage). I will start with very simple implementation and helper functions will be added gradually.

Also as suggested here #47, database call includes `project_id` to avoid reading project first and then release in two separate queries.

bolshakoff (approved):
Nice.

PR #52, opened 2024-05-10
refactor: move auth logic from repository to auth client

jan-zabloudil:
Since Supabase will be used mainly for authentication, `AuthClient` is created to encapsulate this logic. The `AuthClient` can be directly used in the middleware.

Furthermore, the `service/AuthService` no longer contains authenticate method and deals only with authorization. Therefore, the service has been renamed to `AuthorizationService`.


auth/client.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (c *Client) Authenticate(ctx context.Context, token string) (uuid.UUID, error) {
+	user, err := c.client.Auth.User(ctx, token)
+	if err != nil {
+		var errResponse *supabase.ErrorResponse
+		if errors.As(err, &errResponse) {
+			switch errResponse.Code {
+			case http.StatusUnauthorized, http.StatusNotFound: // 404 is returned if valid token is provided but user no longer exists
+				return uuid.UUID{}, fmt.Errorf("%w: %s", ErrInvalidOrExpiredToken, errResponse.Message)
```
jan-zabloudil, line 36: The strategy for error handling for packages that "serve" service is [described here](https://github.com/jan-zabloudil/release-manager/pull/51#discussion_r1597202217). 

However, the `AuthClient` is used in the transport layer. I believe it would be better for the `AuthClient` to return its own specific error. The transport layer could then translate this error into appropriate transport error. 

I think this approach is more effective because the transport layer is a concrete implementation for the REST API. if a different transport layer would be chosen, it can use `AuthClient` as well and translate auth error to its own error.

bolshakoff: Yep, good move. 👍

bolshakoff (approved):

PR #54, opened 2024-05-13
refactor: use db pool in settings repository

jan-zabloudil:


service/settings.go
```diff
@@ -34,16 +34,14 @@ func (s *SettingsService) Update(ctx context.Context, u model.UpdateSettingsInpu
...
 	}
 
-	settings, err := s.Get(ctx, authUserID)
-	if err != nil {
-		return model.Settings{}, err
-	}
+	settings, err := s.repository.Update(ctx, func(s model.Settings) (model.Settings, error) {
+		if err := s.Update(u); err != nil {
+			return model.Settings{}, apierrors.NewSettingsUnprocessableError().Wrap(err).WithMessage(err.Error())
+		}
```
jan-zabloudil, line 40: Reading settings and updating it is done in one transaction.

bolshakoff: Repository returning new `settings` is fine. The rest also looks OK now.

bolshakoff: (optional) Regarding one transaction, good ambition, and the current implementation is probably fine, albeit perhaps too low-level.
But do we really need it? What's the worst case and what are the potential consequences?
Or as a compromise, consider using `SELECT ... FOR UPDATE`.

bolshakoff (approved):

service/settings.go
jan-zabloudil: Thanks 👍 I've decided to try the approach with transactions to practice it. I'm going to keep the current implementation, but I'll add a comment to consider a implementation you suggested. Once I gain more experience while implementing the repository for this app, I might switch to a simpler approach.

PR #56, opened 2024-05-14
refactor: rename service github interface

jan-zabloudil:


bolshakoff (approved):
Client was OK as well, simple and extremely readable, but I understand the intention to have it clean in the sense that service should not care whether it is directly client or a manager wrapper, which is also fine for me.

PR #61, opened 2024-05-14
refactor(repository): use db pool in functionality for deleting project invitation

jan-zabloudil:


repository/project.go
```diff
@@ -117,7 +117,7 @@ func (r *ProjectRepository) DeleteProject(ctx context.Context, id uuid.UUID) err
 	}
 
 	if result.RowsAffected() == 0 {
-		return apierrors.NewProjectNotFoundError().Wrap(err)
+		return apierrors.NewProjectNotFoundError()
```
jan-zabloudil, line 120: Just tweaking changes from previous PR. There is no error to wrap (error is nil that point).

repository/query/query.go
```diff
@@ -31,4 +31,9 @@ var (
 
 	//go:embed scripts/create_project_member.sql
 	CreateProjectMember string
+
+	//go:embed scripts/delete_invitation_for_project.sql
+	DeleteInvitationForProject string
+	//go:embed scripts/delete_project_invitation_by_hash.sql
+	DeleteProjectInvitationByHashAndStatus string
```
bolshakoff, line 38: Let's make the names more consistent.

bolshakoff (approved):

PR #68, opened 2024-05-16
refactor(repository): use db pool for listing project invitations

jan-zabloudil:


service/project.go
```diff
@@ -299,15 +299,22 @@ func (s *ProjectService) ListInvitations(ctx context.Context, projectID, authUse
 		return nil, err
 	}
 
-	exists, err := s.projectExists(ctx, projectID)
+	i, err := s.repo.ListInvitationsForProject(ctx, projectID)
```
bolshakoff, line 302: `i` is a bit too harsh)

repository/query/scripts/list_invitations_for_project.sql
```diff
@@ -0,0 +1,3 @@
+SELECT *
+FROM project_invitations
+WHERE project_id = @projectID
```
bolshakoff, line 3: Here, and in other listing methods, consider adding a sorting criterion (most simple case - `created_at`) so that it is always deterministic.

bolshakoff (approved):

PR #69, opened 2024-05-16
 refactor(repository): use db pool for reading environment

jan-zabloudil:


repository/query/query.go
```diff
@@ -53,4 +53,8 @@ var (
 	ListEnvironmentsForProject string
 	//go:embed scripts/delete_environment.sql
 	DeleteEnvironment string
+	//go:embed scripts/read_environment.sql
+	ReadEnvironment string
+	//go:embed scripts/read_environment_by_name_for_project.sql
+	ReadEnvironmentByNameForProject string
```
jan-zabloudil, line 59: In case of `Members`, `Invitations`, `Environments`..... when reading record by primary key, primary key is composite and contains project_id. 

But I think the query (that is using primary key for filtering) can be simply called `Read{Object}` (instead of `Read{Object}ForProject`. 

But in other cases when primary key is not used, it is better to include "where" conditions in the name -> `ReadEnvironmentByNameForProject`

jan-zabloudil: And the same goes with the function naming, since functions are in `ProjectRepository`. `ReadInvitation` (instead of `ReadInvitationForProject` / `ReadProjectInvitation`) is enough I believe.

I even `ListInvitationsForProject` could be simplified to `ListInvitations`....

Let me know what you think

bolshakoff: Regarding the first one - definitely. 💯 

Regarding the functions names, it's less pronounced, because it's more high-level etc.
So, both options are fine from my side. 🤷‍♀️ 👌

bolshakoff: I was hesitant for a while, but probably you are right.
And the main argument for me is that yes, even on the level of business logic, it is `project` domain, meaning we are inside a project.
So yeah. 👍

repository/mock/project.go
```diff
@@ -44,8 +44,8 @@ func (m *ProjectRepository) CreateEnvironment(ctx context.Context, e svcmodel.En
 	return args.Error(0)
 }
 
-func (m *ProjectRepository) ReadEnvironment(ctx context.Context, envID uuid.UUID) (svcmodel.Environment, error) {
-	args := m.Called(ctx, envID)
+func (m *ProjectRepository) ReadEnvironment(ctx context.Context, projectID, envID uuid.UUID) (svcmodel.Environment, error) {
```
bolshakoff, line 47: As mentioned in a neighbor thread, why do we need `projectID` here?

repository/query/query.go
bolshakoff: `ListEnvironments` in project domain is nice though (which you already have).
And `ReadEnvironmentByName`.

jan-zabloudil: Endpoint is `/projects/{project_id}/environments/{env_id}`

As agreed here #47 - when reading environment, include project_id to catch case when bad project id is provided (but existing env id). 

Before I did it in two queries: 1) read project 2) read env

jan-zabloudil: I think this confusion is a sign that `ForProject` should be included.

jan-zabloudil: Or it would deserve a comment in repository layer

bolshakoff: Yes, sounds fair. 👍

bolshakoff (approved):

PR #76, opened 2024-05-16
feat: add sorting criteria to listing queries

jan-zabloudil:
I think for users list and project members list it makes sense to order records by user's name.

In case of environments you usually want order: dev, test, staging, prod - so I will probably add `order` column to `environment` entity (in separate PR). For now environments are sorted by name as well.


bolshakoff:
I wouldn't add a new column to `environments` db table.
I think it would be better to simply do the ordering on the server (simpler db schema, more explicit logic, more flexible, no significant latency difference obviously).

For the rest - OK. 👍

bolshakoff (approved):

PR #78, opened 2024-05-18
refactor: improve accept invitation logic in service; use db pool in repo

jan-zabloudil:
Just general note:

These are the last few PRs for switching from the Supabase client to the DB pool :tada:. 

Once the final one is merged, I will do one more clean-up PR where I remove all unused code and unify the naming, as there may still be some inconsistencies.

### DIFF
--- a/repository/query/scripts/read_invitation_by_id_and_status_for_update.sql
+++ b/repository/query/scripts/read_invitation_by_id_and_status_for_update.sql
@@ -1,0 +1,6 @@
+SELECT *
+FROM project_invitations
+WHERE
+    id = @id AND
+    status = @status
+FOR UPDATE
--- a/service/project.go
+++ b/service/project.go
@@ -2,10 +2,10 @@ package service
 
 import (
 	"context"
+	"fmt"
 
 	"release-manager/pkg/apierrors"
 	cryptox "release-manager/pkg/crypto"
-	"release-manager/pkg/dberrors"
 	"release-manager/service/model"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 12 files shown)


repository/query/scripts/read_invitation_by_id_and_status_for_update.sql
```diff
@@ -0,0 +1,6 @@
+SELECT *
+FROM project_invitations
+WHERE
+    id = @id AND
+    status = @status
+FOR UPDATE
```
jan-zabloudil, line 6: In the next PR I added a different approach to appending "FOR UPDATE" to the queries.

https://github.com/jan-zabloudil/release-manager/pull/80#pullrequestreview-2064764277

service/project.go
```diff
@@ -338,31 +338,56 @@ func (s *ProjectService) CancelInvitation(ctx context.Context, projectID, invita
...
+	// Then we check if user with given email exists
+	// First two steps could be done in one query, but it would require mixing entities from two different repositories
+	// Based on user existence we either accept the invitation or create a project member and delete the invitation
+	//
+	// There is one possible raise condition:
+	// 1. We check if user exists (and he does not)
+	// 2. We only accept the invitation
+	// But user would register between these two steps
+	// Resulting in a project member not being created for existing user
+	// The raise condition is handled by the PostgreSQL function check_accepted_invitations_for_registered_users()
```
bolshakoff, line 357: Might be a bit overkill, but love the explanation and diligence. 💪 
(also, you probably mean race condition, right?)

bolshakoff (approved):

service/project.go
jan-zabloudil: > (also, you probably mean race condition, right?)

haha, yes 🙈

PR #80, opened 2024-05-18
refactor(repository): use db pool for updating project member

jan-zabloudil:


repository/query/query.go
```diff
@@ -72,3 +72,7 @@ var (
 	//go:embed scripts/read_environment_by_name.sql
 	ReadEnvironmentByName string
 )
+
+func AppendForUpdate(query string) string {
+	return query + "\nFOR UPDATE"
+}
```
jan-zabloudil, line 78: I got tired of writing separate queries just to append 'for update'. Additionally, I believe that this approach is better from a manageability perspective.

bolshakoff (approved):

PR #83, opened 2024-05-19
feat: add listing releases endpoint

jan-zabloudil:


repository/query/scripts/list_releases_for_project.sql
```diff
@@ -0,0 +1,3 @@
+SELECT *
+FROM releases
+WHERE project_id = @projectID
```
bolshakoff, line 3: Sorting

bolshakoff (approved):

PR #84, opened 2024-05-19
feat: add check if release name is unique

jan-zabloudil:


service/release.go
```diff
@@ -50,7 +50,14 @@ func (s *ReleaseService) Create(
...
-	// TODO check if release name is unique per project
+	unique, err := s.isReleaseTitleUniqueInProject(ctx, projectID, rls.ReleaseTitle)
+	if err != nil {
+		return model.Release{}, err
+	}
+	if !unique {
+		return model.Release{}, apierrors.NewReleaseDuplicateTitleError().Wrap(err)
+	}
+
 	if err := s.repo.Create(ctx, rls); err != nil {
```
bolshakoff, line 61: Can't we simply catch an appropriate SQL error here (smth like unique constraint violation).
Maybe already asked, but don't remember the reason.. 🤔
In such case, might deserve to be added as a comment...

bolshakoff (approved):

service/release.go
jan-zabloudil: No, we can. I updated the PR and love the simplicity of the solution 👍

Originally, I thought you should always try to catch the error to avoid reaching the database if possible. However, I can see that this approach (at least in this case) creates more complexity and could result in a race condition, where a new record with the same name might be created in the meantime (after you check the name but before you insert a new record).

If you approve the solution I will also use the solution in the rest of the application.

repository/release.go
```diff
@@ -36,6 +37,10 @@ func (r *ReleaseRepository) Create(ctx context.Context, rls svcmodel.Release) er
 		"updatedAt":    rls.UpdatedAt,
 	})
 	if err != nil {
+		if util.IsUniqueConstraintViolation(err, "unique_release_title_per_project") {
```
jan-zabloudil, line 40: In the case of `releases` entity, a unique constraint violation can occur in two situations:

- Inserting a duplicated primary key (which, in the case of UUIDs, is highly unlikely in the real world of course).
- Inserting a duplicated release title.

Therefore, it would be acceptable to simply catch the unique constraint error without checking the specific constraint and return a `NewReleaseDuplicateTitleError`. However, I believe it is better to make the handling more explicit.

service/release.go
bolshakoff: Yep. Repo returns API error. 👌

repository/release.go
bolshakoff: OK, I like the explicitness.
But then perhaps the name of the constraint deserves to be a constant.

bolshakoff: (or simply returning a more general `ALREADY_EXISTS` error - also fine)

bolshakoff (approved):

PR #86, opened 2024-05-19
feat(github client): add funcionality for creating a github release

jan-zabloudil:
When user creates a new release, **existing*** git tag must be provided in order to link release to source code. Once these PR are merged, release service will:
- check if provided git tag actually exists (https://github.com/jan-zabloudil/release-manager/pull/87)
- check if provided git tag is associated with github release (https://github.com/jan-zabloudil/release-manager/pull/88)
- create a github release (if tag is not associated with one yet) and user allows github release creation (this PR)

---

*) In the first MVP version, only existing git tags will be supported. Later, the functionality will be extended to allow the use of new git tags as well (`ReleaseManager` would create a new git tag).

### DIFF
--- a/github/client.go
+++ b/github/client.go
@@ -29,13 +29,10 @@ func (c *Client) ReadTagsForRepository(ctx context.Context, tkn string, repoURL 
 		return nil, apierrors.NewGithubRepositoryInvalidURL().Wrap(err).WithMessage(err.Error())
 	}
 
-	// GitHub client is created in the function because it needs to be authenticated with a token (that is passed as an argument)
-	client := github.NewClient(nil).WithAuthToken(tkn)
-
 	// Up to 100 tags can be fetched per page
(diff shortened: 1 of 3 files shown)


pkg/apierrors/apierrors.go
```diff
@@ -205,6 +206,13 @@ func NewGithubRepositoryNotFoundError() *APIError {
 	}
 }
 
+func NewGithubReleaseCreationError() *APIError {
```
bolshakoff, line 209: What category is it? )

bolshakoff (approved):

pkg/apierrors/apierrors.go
jan-zabloudil: Good point. I have realized I don't yet know if the action will be executed asynchronously. So, for now, I will just return the error. Once I know how the function is used in the service, I will handle the service error accordingly.

PR #87, opened 2024-05-19
feat(github client): add functionality for reading a git tag

jan-zabloudil:
When user creates a new release, **existing*** git tag must be provided in order to link release to source code. Once these PR are merged, release service will:
- check if provided git tag actually exists (this PR)
- check if provided git tag is associated with github release (https://github.com/jan-zabloudil/release-manager/pull/88)
- create a github release (if tag is not associated with one yet) and user allows github release creation (https://github.com/jan-zabloudil/release-manager/pull/86)

---

*) In the first MVP version, only existing git tags will be supported. Later, the functionality will be extended to allow the use of new git tags as well (`ReleaseManager` would create a new git tag).


pkg/apierrors/apierrors.go
```diff
@@ -240,6 +241,13 @@ func NewSlackIntegrationNotEnabledError() *APIError {
 	}
 }
 
+func NewGitTagNotFoundError() *APIError {
```
bolshakoff, line 244: Doesn't it belong to NotFoundError?

bolshakoff (approved):

PR #89, opened 2024-05-20
feat: add project invitation email template

jan-zabloudil:
Snímek obrazovky 2024-05-20 v 13 20 17

The email template is simple, but I believe it is sufficient for the MVP.

I also think that storing and parsing email templates on the backend is not a very maintainable solution in the case of Go. However, it should be sufficient for the MVP.

The email template was specifically tested for Gmail.


resend/templates/project_invitation.tmpl
```diff
@@ -0,0 +1,41 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+</head>
+<body>
+    <table class="email-container" cellpadding="0" cellspacing="0" width="100%" style="font-family: Arial, sans-serif; max-width: 600px; margin: 0 auto; background-color: #ffffff; padding: 20px; border-radius: 8px; box-shadow: 0 0 10px rgba(0, 0, 0, 0.1);">
+        <tr>
+            <td>
+                <h1 style="text-align: center; font-size: 24px; margin-bottom: 20px;">Project Invitation</h1>
+                <p style="font-size: 16px; line-height: 1.5;">Hey,</p>
```
bolshakoff, line 27: "Hey," feels weirdly informal and at the same time cold. 😄 🥶 

Consider removing the salutation altogether, as it's hard to get right.

No salutation is fine, professional and quite user-friendly, as it gets to the point, which in case of generated messages a good thing, I would say.

bolshakoff (approved):
Nice.

PR #91, opened 2024-05-21
feat: add deployed release to environment

jan-zabloudil:
To keep this PR smaller, it only updates the read functionality. Create and update functionality will be updated once this PR is merged

### DIFF
--- a/repository/model/environment.go
+++ b/repository/model/environment.go
@@ -1,6 +1,8 @@
 package model
 
 import (
+	"database/sql"
+	"fmt"
 	"net/url"
 	"time"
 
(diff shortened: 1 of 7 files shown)


repository/model/environment.go
```diff
@@ -9,44 +11,60 @@ import (
...
 type Environment struct {
-	ID         uuid.UUID `json:"id" db:"id"`
-	ProjectID  uuid.UUID `json:"project_id" db:"project_id"`
-	Name       string    `json:"name" db:"name"`
-	ServiceURL string    `json:"service_url" db:"service_url"`
-	CreatedAt  time.Time `json:"created_at" db:"created_at"`
-	UpdatedAt  time.Time `json:"updated_at" db:"updated_at"`
-}
-
-type UpdateEnvironmentInput struct {
```
jan-zabloudil, line 22: Unused code from repository model removed.

bolshakoff (commented):
At first sight, looks very nice.
But when I think of it:
1. Given that in wireframes the pivot of deployments is release, I would expect that here, also, deployments would be a part of release.
2. Or maybe wireframes / app vision deserves to be enriched with a view at environments. Currently, it's just links to environments. (Or I'm missing something.)
3. We might end up prioritizing both views: Releases x where it is deployed AND Environments x deployed releases. In this case also, it would be great to see it already here, in this PR - how would the models look like and perhaps Read operation (btw, it's great that you split it like that 💪 ).
4. And/or actually having sort of a 3rd view - deployments - where you simply see a history of what was deployed where and when. Apart from being a nice, useful view in general, might be useful for debugging (for example, I'm frontend and yesterday I got errors - was it because that release contained errors or because it wasn't deployed yet).
5. And even if we don't go for that "3rd, deployments view" in this MVP, at least on the data layer it might be the best, the most flexible thing to have, i.e. `deployments` table where simply all deployments (environment ID x release ID) are recorded. And then anything can be reconstructed from that, albeit with slightly more complicated queries.
6. One of the things that motivates me now is that when I read the code now, I constantly find myself in a confusion like "wait, deployment environment is a part of release or deployed release is a part of environment"...

jan-zabloudil:
Thank you! I need to reconsider this, so I am closing this PR for now.

PR #94, opened 2024-05-22
refactor: move response errors to transport package; response errors translate api errors

jan-zabloudil:
- Response errors were moved to `transport` package
- `responseerrors` package is now responsible for translating api errors.
- next step is to move api errors to service package (-> service errors). It will be done in separate PR once this one is merged.


transport/responseerrors/errormapper.go
```diff
@@ -0,0 +1,77 @@
+package responseerrors
```
jan-zabloudil, line 1: To avoid import conflicts with the `errors` package, the package is still named `responseerrors`.

transport/responseerrors/responseerrors.go
```diff
@@ -24,116 +24,116 @@ var (
 	errCodeUnknown                    = "ERR_UNKNOWN"
 )
 
-type ResponseError struct {
+type Error struct {
```
jan-zabloudil, line 27: Error struct can be simply named `Error`.

pkg/apierrors/apierrors.go
```diff
@@ -63,262 +63,229 @@ func (e *APIError) WithMessage(msg string) *APIError {
...
-		IsErrorWithCode(err, errCodeGithubIntegrationNotEnabled) ||
-		IsErrorWithCode(err, errCodeProjectMemberNotFound) ||
-		IsErrorWithCode(err, errCodeGithubIntegrationNotEnabled) ||
-		IsErrorWithCode(err, errCodeGithubRepositoryInvalidURL) ||
-		IsErrorWithCode(err, errCodeReleaseNotFound) ||
-		IsErrorWithCode(err, errCodeGitTagNotFound) ||
-		IsErrorWithCode(err, errCodeGithubReleaseNotFound)
-}
-
-func IsUnprocessableModelError(err error) bool {
```
jan-zabloudil, line 292: `apierrors` now has only `IsXXXError(err error) bool` methods that are used in the service.

pkg/apierrors/apierrors.go
```diff
@@ -7,35 +7,35 @@ import (
...
-	errCodeGithubRepositoryInvalidURL              = "ERR_GITHUB_REPOSITORY_INVALID_URL"
-	errCodeProjectMemberNotFound                   = "ERR_PROJECT_MEMBER_NOT_FOUND"
-	errCodeProjectMemberUnprocessable              = "ERR_PROJECT_MEMBER_UNPROCESSABLE"
-	errCodeReleaseUnprocessable                    = "ERR_RELEASE_UNPROCESSABLE"
-	errCodeReleaseNotFound                         = "ERR_RELEASE_NOT_FOUND"
-	errCodeReleaseDuplicateTitle                   = "ERR_RELEASE_DUPLICATE_TITLE"
-	errCodeSlackIntegrationNotEnabled              = "ERR_SLACK_INTEGRATION_NOT_ENABLED"
-	errCodeGitTagNotFound                          = "ERR_GIT_TAG_NOT_FOUND"
-	errCodeGithubReleaseNotFound                   = "ERR_GITHUB_RELEASE_NOT_FOUND"
+	ErrCodeUnauthorizedUnknownUser                 = "ERR_UNAUTHORIZED_ACCESS_UNKNOWN_USER"
```
jan-zabloudil, line 10: `apierrors` error codes are now public, so `responseerrors` can translate the errors into appropriate response errors based on the code.

transport/responseerrors/errormapper.go
jan-zabloudil: I am also considering shorter name `resperrors` to avoid double e - `responsEEerrors`. What do you think @bolshakoff ?

bolshakoff: Since `transport` and `service` are different packages, there shouldn't be a conflict, and I'd have the same name for them.
And during import yes, they will have aliases like `apierrors` and `resperrors`.

bolshakoff (approved):
I like the direction. Just one comment about import aliases.

PR #99, opened 2024-05-23
refactor(trasport): remove unnecessary setting of time variables to local time

jan-zabloudil:
~~Added for `project_member` and `release`, for other models already added earlier.~~


bolshakoff (approved):
Hmm, can you describe more what is the motivation behind this change? 🤔

jan-zabloudil:
> Hmm, can you describe more what is the motivation behind this change? 🤔

Initially, I had issues displaying time in the correct time zone. However, since the time is saved as `TIMESTAMP WITH TIME ZONE`, it works fine without using `.Local()`. Therefore, I removed it.

bolshakoff (approved):
OK 😄

PR #102, opened 2024-05-25
feat(slack client): add synchronous message sending functionality

jan-zabloudil:


slack/client.go
```diff
@@ -35,25 +49,37 @@ func (c *Client) SendReleaseNotificationAsync(ctx context.Context, token, channe
...
+func (c *Client) sendMessage(ctx context.Context, token, channelID string, msgOptions []slack.MsgOption) error {
+	client := slack.New(token)
+
+	if _, _, err := client.PostMessageContext(ctx, channelID, msgOptions...); err != nil {
+		switch err.Error() {
+		case errInvalidAuth:
+			return svcerrors.NewSlackClientUnauthorizedError().Wrap(err)
+		case errChannelNotFound:
+			return svcerrors.NewSlackChannelNotFoundError().Wrap(err).
+				WithMessage(fmt.Sprintf("slack channel (ID: %s) not found", channelID))
```
bolshakoff, line 78: Will these errors be caught and translated to transport errors?

jan-zabloudil: Yes, once I merge this one and #102. I will use this function in the service and also add translation to transport errors.

bolshakoff: Yes, but since it runs asynchronously, I presume it will only be caught by background manager and only logged, no?.. 🤔

jan-zabloudil: In case of async function, error is logged by task manager

```
t := task.Task{
  Type: task.TypeOneOff,
  Meta: task.Metadata{
	  "task": "sending slack message",
  },
  Fn: func(ctx context.Context) error { <--- HERE
	  return c.sendMessage(ctx, token, channelID, msgOptions)
  },
}
```

Log snippet:

```
{"time":"2024-05-27T15:33:19.608265+02:00","level":"ERROR","msg":"task:failed","task":{"type":"oneoff","meta":{"task":"sending slack message"}},"error":"Code: ERR_SLACK_CHANNEL_NOT_FOUND, error: channel_not_found"}
```

jan-zabloudil: Or you dont like the fact that in case of async function, the function also returns service error (that is then logged by task manager)?

bolshakoff (approved):

slack/client.go
jan-zabloudil: We will use both versions.

- When creating a release, Slack message will be sent async.
- But we will also have [dedicated endpoint](https://github.com/jan-zabloudil/release-manager/pull/102) to send Slack message later on (when release is already created). 
- And here we want to have sync function - as suggested [here](https://www.notion.so/release-manager/Separate-action-for-sending-Slack-notification-20a137f02d19487d9b3cc991422175ce). 
- For dedicated endpoint it is good to have sync version so we are able to show error to user when operation fails and user can find out what went wrong with the integration setup (wrong channel id, token....)

PR #103, opened 2024-05-25
feat: add endpoint for sending release notification

jan-zabloudil:


transport/handler/routes.go
```diff
@@ -67,6 +67,7 @@ func (h *Handler) setupRoutes() {
 					r.Use(middleware.HandleResourceID("release_id", util.ContextSetReleaseID))
 					r.Get("/", middleware.RequireAuthUser(h.getRelease))
 					r.Patch("/", middleware.RequireAuthUser(h.updateRelease))
+					r.Post("/notify", middleware.RequireAuthUser(h.sendReleaseNotification))
```
jan-zabloudil, line 70: In the end I used more action oriented route `/projects/{project_id}/releases/{release_id}/notify`

transport/errors/errormapper.go
```diff
@@ -67,6 +69,11 @@ func isConflictError(err error) bool {
 		isSvcErrorWithCode(err, svcerrors.ErrCodeReleaseDuplicateTitle)
 }
 
+func isBadRequestError(err error) bool {
+	return isSvcErrorWithCode(err, svcerrors.ErrCodeSlackChannelNotSetForProject) ||
+		isSvcErrorWithCode(err, svcerrors.ErrCodeSlackIntegrationNotEnabled)
```
jan-zabloudil, line 74: If user wants to send a release notification and slack integration is either not enabled or no slack channel is set for project, `bad request` is returned. Let me know if you would rather choose different http status code.

service/release.go
```diff
@@ -139,7 +139,37 @@ func (s *ReleaseService) ListForProject(ctx context.Context, projectID, authorUs
...
+		return svcerrors.NewSlackChannelNotSetForProjectError()
+	}
+
+	tkn, err := s.settingsGetter.GetSlackToken(ctx)
+	if err != nil {
+		return fmt.Errorf("getting slack token: %w", err)
+	}
+
+	// TODO use synchronous notification sending
+	s.slackNotifier.SendReleaseNotificationAsync(ctx, tkn, p.SlackChannelID, model.NewReleaseNotification(p, rls))
```
jan-zabloudil, line 167: The synchronous send release notification method will be implemented in a separate PR.

jan-zabloudil: https://github.com/jan-zabloudil/release-manager/pull/102

transport/handler/routes.go
bolshakoff: After a quick search, couldn't find the official docs, but it seems like there is a strong consensus in the community to stick to nouns.
Would love to find the opposite evidence, though.

And while now it might seem too artificial to name it smth like `POST notification`, it actually might be a persistent resource later on (for example, we might decide to store these notification triggers for analytics purposes).

transport/errors/errormapper.go
bolshakoff: Yeah, I also don't see a better option. 😕

bolshakoff (commented):
I think we can and should stick to the nouns in endpoint names.

transport/handler/routes.go
jan-zabloudil: Ok, I would choose `/slack-notification`

bolshakoff: Wow, I can edit other people's comments 😄 

I just wanted to add - let's be 100% RESTful and go with plural. 👌

jan-zabloudil: Ok, good point. Thanks!

If there is no other think you would change, please approve it :) thank you
I will update the route before merging.

bolshakoff (approved):

PR #104, opened 2024-05-25
feat(transport): add endpoint to handle github webhook requests

jan-zabloudil:


transport/handler/webhook.go
jan-zabloudil, line 1: If I had more routes for handling GitHub webhooks, I would move the logic for verifying the signature to middleware. However, since there is only one endpoint for now, I keep the whole logic in the handler.

Since I need to read the entire body:
```
body, err := io.ReadAll(r.Body)
```

and then unmarshal it later:
```
if err := json.Unmarshal(body, &input); err != nil {
```

it is more convenient to do both in one function. Otherwise, I would have to pass the body in the context (because you can call `io.ReadAll(r.Body)` only once).

transport/handler/webhook.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (h *Handler) handleGithubReleaseWebhook(w http.ResponseWriter, r *http.Request) {
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		util.WriteResponseError(w, resperrors.NewBadRequestError().Wrap(err))
+		return
+	}
+
+	// TODO get the secret from the settings
+	secret := ""
```
jan-zabloudil, line 26: https://github.com/jan-zabloudil/release-manager/pull/105

bolshakoff (approved):
Nice.

PR #108, opened 2024-05-25
feat: require git tag when creating release

jan-zabloudil:


bolshakoff:
One thing that came to my mind now:
- We allow having GitHub integration disabled in `settings`, but when creating a project, GitHub URL cannot be empty.

Might be an area for refinement.

jan-zabloudil:
> We allow having GitHub integration disabled in `settings`, but when creating a project, GitHub URL cannot be empty.

Good point, thank you. It will be solved as part of "Github" refactoring 👍

PR #105, opened 2024-05-25
feat(settings): add github webhook secret

jan-zabloudil:

PR #109, opened 2024-05-26
Feature/GitHub release

jan-zabloudil:
- Updates the `GithubRelease` model in the service
- Updates the release creation process in the repository to also save GitHub release data


github/model/model.go
```diff
@@ -51,19 +52,19 @@ func ToSvcGitTags(tags []*github.RepositoryTag) []svcmodel.GitTag {
...
-		TagName:     release.GetTagName(),
-		CreatedAt:   release.CreatedAt.Time,
-		PublishedAt: release.PublishedAt.Time,
+		ID:             rls.GetID(),
+		OwnerSlug:      repo.OwnerSlug,
+		RepositorySlug: repo.RepositorySlug,
+		GitTagName:     rls.GetTagName(),
+		HTMLURL:        *u,
+		CreatedAt:      rls.CreatedAt.Time,
+		UpdatedAt:      time.Now(),
```
jan-zabloudil, line 68: To keep things simple, I only store the necessary information in the GithubRelease model.

Since the `GithubRelease` name and body will be synced with `Release.ReleaseTitle `and `Release.ReleaseNotes`, it doesn't make sense to also save name and body in `GithubRelease`.

supabase/migrations/20240505115913_releases.sql
```diff
@@ -6,7 +6,12 @@ CREATE TABLE public.releases (
      created_by UUID REFERENCES public.users ON DELETE SET NULL,
      created_at TIMESTAMP WITH TIME ZONE NOT NULL,
      updated_at TIMESTAMP WITH TIME ZONE NOT NULL,
-     CONSTRAINT unique_release_title_per_project UNIQUE (project_id, release_title)
+     github_release_id BIGINT NOT NULL,
+     github_owner_slug TEXT NOT NULL,
+     github_repo_slug TEXT NOT NULL,
+     github_release_data JSON NOT NULL,
+     CONSTRAINT unique_release_title_per_project UNIQUE (project_id, release_title),
+     CONSTRAINT unique_github_release_per_project UNIQUE (project_id, github_owner_slug, github_repo_slug, github_release_id)
```
jan-zabloudil, line 14: Constraint to make sure that the same release is not used twice in the same project.

---

You really feel the need for refactoring here to have separate `github_repo` and `github_releases` tables. 

This will also make it easier to check if a single GitHub repository is not used in more than one project. In current implementation more projects could connect to the same github repository.

bolshakoff: Also, circling back to our discussion on Friday:
1. If a GitHub Release is created before release here, and release title here is different from GitHub Release title - what do we do? I guess we missed that scenario on Friday, but the solution is obvious - update it with the user input, right?..
2. Actually, since we are here, in general, what was the main reason for keeping release titles in sync (between release manager and GitHub)? Can't we allow them to be different? Maybe even.. we want them to be decoupled, for the following reasons. One, if I only use, only know release manager, I would be unpleasantly surprised if today I have one release title, and tomorrow another one, even though I didn't change it (but smb changed it in GitHub). Two, if I'm already using a dedicated release manager, I don't want to bother with GitHub at all, and I'm not gonna manipulate releases there.

Continuing 2, let's think about simplifying the program even more.
Because as a user, again, if I use Release Manager, what are the scenarios:
- I create releases only via Release Manager. I read releases only via Release Manager. I update releases only via Release Manager (for example, if I forgot to say in release notes that ticket XXX is also included into it). So, **I never update or delete releases via GitHub. In fact, I don't even need it to be created in GitHub. I don't care about GitHub at all.**
- I create releases in GitHub, because I have a CI/CD workflow that does that automatically after I push a new tag. That release contains a nice changelog, compared to the previous release (that's actually what I had on my project 😄). I never modify it. I then go to Release Manager and trigger release creation there. Aha. Rather than a potential checkbox "Also create a GitHub Release" **I'd rather need a checkbox or even a separate button "GitHub Release already created" which will require only git tag from me, and release title and release body will be copied from GitHub**. Now that the release is in Release Manager, similar to the previous workflow, **I never update or delete it via GitHub**. And if I want to add a couple sentences to release in Release Manager, I can do it, and I will definitely do it there, **and I don't even want it to affect GitHub Release. GitHub Release will stay sort of formal and dry there, forever.**

So, I'd say we might not need the hooks at all.

And the same in the opposite direction - if I update release in Release Manager, I don't want it to affect release in GitHub.

And when creating a release, I either want to do it fully via Release Manager (and I don't care about GitHub), or I want to create it from an already created GitHub Release (because I have a process that generates nice releases in GitHub automatically, on git tag push).

github/model/model.go
bolshakoff: Feels a bit too complicated. 😄 

Related to a huge comment below (or above), from a usage perspective, do we need any synchronization with GitHub Releases at all? Even the link to GitHub Release is unnecessary in both scenarios.

bolshakoff (commented):
Things are getting too complicated for my lazy head 😄 hence a proposition to simplify things even more.

bolshakoff:
Oh no, from my view it looked nice, and like in one thread one by one.. 🤦‍♀️

bolshakoff:
But actually OK, it's in chronological order from top to bottom. 👌

bolshakoff:
And it's duplicated. 😶 🤦‍♀️

bolshakoff:
Blocks #105 for a while.

jan-zabloudil:
After discussions outside of GitHub, we concluded that we will maintain integration with GitHub. However, we will simplify the service and repository GitHub release model and remove the denormalization state created by keeping GitHub repository slugs in the GitHub release repo model.

service/model/release.go
```diff
@@ -109,11 +109,10 @@ func NewReleaseNotification(p Project, r Release) ReleaseNotification {
...
-	HTMLURL     url.URL // URL to the release page on GitHub
-	CreatedAt   time.Time
-	PublishedAt time.Time
+	GitTagName string
+	// URL to the release page on GitHub.
+	HTMLURL   url.URL
+	CreatedAt time.Time
+	// Time when the up-to-date information was last fetched from GitHub.
+	UpdatedAt time.Time
 }
```
jan-zabloudil, line 118: `GitHub` model simplified.

supabase/migrations/20240505115913_releases.sql
```diff
@@ -5,7 +5,11 @@ CREATE TABLE public.releases (
      release_notes TEXT,
      created_by UUID REFERENCES public.users ON DELETE SET NULL,
      created_at TIMESTAMP WITH TIME ZONE NOT NULL,
-     updated_at TIMESTAMP WITH TIME ZONE NOT NULL
+     updated_at TIMESTAMP WITH TIME ZONE NOT NULL,
+     git_tag_name Text NOT NULL,
+     github_release_created_at TIMESTAMP WITH TIME ZONE NOT NULL,
+     github_release_published_at TIMESTAMP WITH TIME ZONE NOT NULL,
+     CONSTRAINT unique_git_tag_per_project UNIQUE (project_id, git_tag_name)
```
jan-zabloudil, line 12: I realized that it can be still easily checked if a Git tag was already used in the project. Since the whole project is connected to one repository, there's no need to check the GitHub repository owner or repository slug. You can simply use the following constraint:

```
 CONSTRAINT unique_git_tag_per_project UNIQUE (project_id, git_tag_name)
```

NOTE:
_Once the github refactoring will be done, you wont'be able to switch your project to different github repository (once you create at least one release). If your project would use different repo, you will need to archive the current one and create a new one._

service/service.go
```diff
@@ -84,6 +84,9 @@ type projectGetter interface {
 
 type githubManager interface {
 	ReadTagsForRepository(ctx context.Context, token string, repoURL url.URL) ([]model.GitTag, error)
+	ReadTagByName(ctx context.Context, token string, repoURL url.URL, tagName string) (model.GitTag, error)
+	CreateRelease(ctx context.Context, token string, repoURL url.URL, input model.CreateReleaseInput) (model.GithubRelease, error)
+	ReadReleaseByTag(ctx context.Context, token string, repoURL url.URL, tagName string) (model.GithubRelease, error)
```
jan-zabloudil, line 89: If https://github.com/jan-zabloudil/release-manager/pull/114 is accepted, functions will accept `svcmodel.GithubRepo` instead of `repoURL`.

supabase/migrations/20240505115913_releases.sql
bolshakoff: Nice. But do we really need the `github_release_created_at` and `github_release_published_at` fields?
(Btw, isn't it supposed to be `github_release_created_at` and `github_release_updated_at`?)

If somebody updates a release in GitHub, then both `updated_at` and `github_release_updated_at` will be updated, right?
If somebody updates a release via Release Manager, well, again, both fields will have to be updated, no?.. 🤔 

If somebody needs finer details on whether it was release in Release Manager who triggered updates or release in GitHub, well he or she can take a look at timestamps in GitHub.

bolshakoff: Oh, and btw, in the scenario when we update release in Release Manager and propagate it to GitHub Releases, this will trigger the webhook and.. infinite loop?.. 💀 🔁 💣

service/model/release.go
```diff
@@ -109,11 +109,10 @@ func NewReleaseNotification(p Project, r Release) ReleaseNotification {
...
-	TagName     string  // Git tag associated with the release
-	HTMLURL     url.URL // URL to the release page on GitHub
-	CreatedAt   time.Time
-	PublishedAt time.Time
+	GitTagName string
+	// URL to the release page on GitHub.
+	HTMLURL   url.URL
+	CreatedAt time.Time
+	// Time when the up-to-date information was last fetched from GitHub.
+	UpdatedAt time.Time
```
bolshakoff, line 117: Same here; do we really need these?
(See comment to db migration for detailed reasoning.)

service/model/release.go
```diff
@@ -109,11 +109,10 @@ func NewReleaseNotification(p Project, r Release) ReleaseNotification {
...
-	ID          int64 // GitHub release ID
-	Name        string
-	Body        string
-	TagName     string  // Git tag associated with the release
-	HTMLURL     url.URL // URL to the release page on GitHub
-	CreatedAt   time.Time
-	PublishedAt time.Time
+	GitTagName string
+	// URL to the release page on GitHub.
+	HTMLURL   url.URL
```
bolshakoff, line 114: `HTMLURL` name feels a bit too clunky..
Is it because `URL` is occupied? In this case, we can simply alias `url` package by smth like `urlx`.

service/model/release.go
```diff
@@ -26,14 +26,14 @@ type UpdateReleaseInput struct {
...
+	ProjectID     uuid.UUID
+	ReleaseTitle  string
+	ReleaseNotes  string
+	AuthorUserID  uuid.UUID
+	CreatedAt     time.Time
+	UpdatedAt     time.Time
+	GithubRelease GithubRelease
 }
 
 func NewRelease(input CreateReleaseInput, projectID, authorUserID uuid.UUID) (Release, error) {
```
bolshakoff, line 39: In case we stick to having `GithubRelease` as a separate type (I'm more inclined to rather flattening it out), is its creation going to be in a separate PR?..

supabase/migrations/20240505115913_releases.sql
jan-zabloudil: > Nice. But do we really need the github_release_created_at and github_release_published_at fields?

You are right. We dont really need it now. Let's drop it.

We could then simply have 

```
type GithubRelease struct {
  GitTagName string
  URL url.URL
}  
```

service/model/release.go
jan-zabloudil: As answered above, let's drop it for now.

service/model/release.go
jan-zabloudil: We can use `URL`. I was just following github naming but you are right it can be simpler.

service/model/release.go
jan-zabloudil: Yes, I would do it in separate PR.

supabase/migrations/20240505115913_releases.sql
jan-zabloudil: Or maybe even better, simply

```
GitTagName string
GithubReleaseURL url.URL
```

bolshakoff: Yes, that's what I meant by "flattening". 👍 

I would prefer it.
Of course, service model doesn't have to reflect the db model 1-to-1, but from practical perspective, it might be much easier to keep it similar for now, and do the splitting later on both svc and db layers.

PR #112, opened 2024-05-27
build: enable app to run in Docker

jan-zabloudil:
This is a basic Docker setup. Later on, it can be tailored for specific clouds (e.g., Google Cloud).

Once you approve, I will also update the README. Since I am not a very experienced Docker user, I would appreciate any feedback or suggestions you might have. Thank you


bolshakoff (approved):
Lets Go Nba

PR #113, opened 2024-05-27
feat(github-client): add function to dynamically generate GitHub release URL

jan-zabloudil:
The GitHub release URL will not be saved in the database. Therefore, I prepared a function in the GitHub client that will dynamically generate the GitHub release URL. 

This function will be called from the service layer. In the service layer, the project model holds only the repository URL, so the repository URL and the tag name are passed to the function.


github/client.go
```diff
@@ -143,6 +143,16 @@ func (c *Client) ReadReleaseByTag(ctx context.Context, tkn string, repoURL url.U
 	return model.ToSvcGithubRelease(rls)
 }
 
+// GenerateGithubReleaseURL generates a URL to the release page on GitHub for a given repository and tag name
+func (c *Client) GenerateGithubReleaseURL(repoURL url.URL, tagName string) (string, error) {
```
jan-zabloudil, line 147: If https://github.com/jan-zabloudil/release-manager/pull/114 is merged, I will update it to accept `svcmodel.GithubRepo`

bolshakoff (approved):

PR #114, opened 2024-05-28
feat: add github repo struct to project service model

jan-zabloudil:
I have realized it will be much more convenient to have a `GithubRepo` struct (which will contain parsed slugs) in the service model (and consequently in the repository model).

Why? Because when handling webhook requests, slugs will be provided as the repository identifier. This will simplify the process, i could directly do:

```
WHERE owner_slug = @ownerSlug AND repo_slug = @repoSlug
```

In this PR, I only demonstrate how it would be implemented. If you like this approach, I will also update the repository model and modify the GitHub client methods to accept `svcmodel.GithubRepo` instead of URLs, etc.

I believe this is also a good step towards the larger GitHub refactoring. After the refactoring, you will have a separate entity for the GitHub repository identifier (such as slugs etc.)


service/model/project.go
```diff
@@ -22,10 +22,17 @@ type Project struct {
 	SlackChannelID            string
 	ReleaseNotificationConfig ReleaseNotificationConfig
 	GithubRepositoryURL       url.URL
+	GithubRepo                GithubRepo
 	CreatedAt                 time.Time
 	UpdatedAt                 time.Time
 }
 
+type GithubRepo struct {
+	HTMLURL   url.URL
```
jan-zabloudil, line 31: I would also keep the URL in the struct because the user will provide a URL. When passing the project back to the transport layer, it will be good to also return the URL so the client can show it to the user (as they won't care about slugs).

service/model/project.go
```diff
@@ -22,10 +22,17 @@ type Project struct {
 	SlackChannelID            string
 	ReleaseNotificationConfig ReleaseNotificationConfig
 	GithubRepositoryURL       url.URL
+	GithubRepo                *GithubRepo
```
jan-zabloudil, line 25: Optional field -> pointer

github/client.go
```diff
@@ -24,6 +24,15 @@ func NewClient() *Client {
 	return &Client{}
 }
 
+func (c *Client) GetGithubRepoFromRawURL(rawURL string) (svcmodel.GithubRepo, error) {
```
jan-zabloudil, line 27: After the GitHub refactoring, this method will call a GitHub API to check if the repository exists and is accessible with the provided token.

bolshakoff (commented):
I buy the argument and ambition to have `GithubRepo` as a separate type and the direction that it will have separate tables later on.
Another argument is to be consistent with the approach we are about to have in case of releases, with `GithubRelease` type.

But wouldn't it be even more nice and consistent to then also have `GithubRepo` on the db layer with the "keys", i.e. `owner_slug` and `repo_slug` and generate `repo_url` dynamically, as is the case with `GithubRelease` type?..

UPDATE: Ah, sorry, I can see that this is exactly your intention for future PR. OK. 💪 🙇‍♂️

bolshakoff (approved):

bolshakoff:
But in #109 I can see that chances are, in case of `Release`, GitHub Release fields might end up simply a part of `Release` model, and splitted out later on, together with db layer.

In this case, consider employing the same principle also here, i.e. `GithubOwnerSlug`, `GithubRepoSlug` and (potentially) dynamically generated `GithubRepoURL` simply as a part of `Project` model. 👌

PR #115, opened 2024-05-28
feat: add tag name to GithubReleaseWebhook input

jan-zabloudil:


transport/model/webhook.go
```diff
@@ -3,12 +3,13 @@ package model
 type GithubReleaseWebhookInput struct {
 	Action  string `json:"action" required:"true"`
 	Release struct {
-		Name string `json:"name" required:"true"`
-		Body string `json:"body"`
+		TagName string `json:"tag_name" required:"true"`
```
jan-zabloudil, line 6: Forgot to add `TagName` which will be used as a release identifier.

transport/handler/webhook.go
```diff
@@ -36,6 +37,11 @@ func (h *Handler) handleGithubReleaseWebhook(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
+	if err := validator.Validate.Struct(input); err != nil {
+		util.WriteResponseError(w, resperrors.NewBadRequestError().Wrap(err))
+		return
+	}
```
jan-zabloudil, line 43: I mostly prefer to do validation in the service layer. But in the case of webhook, I feel it makes sense to validate webhook request in the transport layer (since webhook request is specific implementation over HTTP) and fail fast if webhook sends invalid payload. 

And then service would expect specific params, instead of accept some general `svcmodel.GithubReleaseWebhookInput`. Service shouldn't care at all about webhooks...

```
DeleteForGithubRelease(ctx contex.Context, ownerSlug, repoSlug, gitTag string) error
```

PR #117, opened 2024-05-28
feat: add dedicated endpoint for setting github repository for a project

jan-zabloudil:
I realized it won't take much work to prepare a dedicated endpoint for setting a GitHub repository for a project. This will make the overall implementation more solid.

I have added a dedicated endpoint to serve this purpose:

`/projects/{id}/github-repository` (singular, as you can only set one repository)

The service calls a GitHub client that verifies your access to the specified GitHub repository (and that the repository actually exists). This is another step towards the GitHub refactoring.

The final step will be to create a separate table for GitHub repositories. For now, I have added the fields to the project service model.

I have also kept the current `project.GithubRepositoryURL` field. I will remove it in a subsequent PR once this one is merged to keep the PR size manageable.

The flow is as follows:
- You create your project (without entering github repo url); the transport project model does not contain the GitHub repository URL.
- You have a dedicated endpoint for setting the connection to the GitHub repository.
- I have implemented only the POST endpoint. If you agree with the implementation, I will also add the GET endpoint.

---

Once this PR is merged, I will also update github client functions (such as reading git tags, creating github release etc.) to accept `svcmodel.GithubRepo` instead of repoURL.

---

If you approve, I wll also add tests.

### DIFF
--- a/repository/model/project.go
+++ b/repository/model/project.go
@@ -1,6 +1,8 @@
 package model
 
 import (
+	"database/sql"
+	"fmt"
 	"net/url"
 	"time"
 
@@ -14,7 +16,10 @@ type Project struct {
 	Name                      string                    `db:"name"`
 	SlackChannelID            string                    `db:"slack_channel_id"`
 	ReleaseNotificationConfig ReleaseNotificationConfig `db:"release_notification_config"`
-	GithubRepositoryURL       string                    `db:"github_repository_url"`
+	GithubRepositoryURL       string                    `db:"github_repository_url"` // TODO remove this field
+	GithubOwnerSlug           sql.NullString            `db:"github_owner_slug"`
+	GithubRepoSlug            sql.NullString            `db:"github_repo_slug"`
+	GithubRepoURL             sql.NullString            `db:"github_repo_url"`
 	CreatedAt                 time.Time                 `db:"created_at"`
 	UpdatedAt                 time.Time                 `db:"updated_at"`
 }
@@ -34,12 +39,27 @@ func ToSvcProject(p Project) (svcmodel.Project, error) {
 		return svcmodel.Project{}, err
 	}
 
+	var githubRepo *svcmodel.GithubRepo
+	if p.GithubOwnerSlug.Valid && p.GithubRepoSlug.Valid && p.GithubRepoURL.Valid {
+		u, err := url.Parse(p.GithubRepoURL.String)
+		if err != nil {
+			return svcmodel.Project{}, fmt.Errorf("failed to parse github repository URL: %w", err)
+		}
+
+		githubRepo = &svcmodel.GithubRepo{
+			URL:       *u,
+			OwnerSlug: p.GithubOwnerSlug.String,
+			RepoSlug:  p.GithubRepoSlug.String,
+		}
+	}
+
 	return svcmodel.Project{
 		ID:                        p.ID,
 		Name:                      p.Name,
 		SlackChannelID:            p.SlackChannelID,
 		ReleaseNotificationConfig: svcmodel.ReleaseNotificationConfig(p.ReleaseNotificationConfig),
-		GithubRepositoryURL:       *u,
+		GithubRepositoryURL:       *u, // TODO remove
+		GithubRepo:                githubRepo,
 		CreatedAt:                 p.CreatedAt,
 		UpdatedAt:                 p.UpdatedAt,
 	}, nil
--- a/repository/project.go
+++ b/repository/project.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
 
 	"release-manager/pkg/crypto"
 	"release-manager/repository/model"
@@ -49,7 +50,7 @@ func (r *ProjectRepository) CreateProjectWithOwner(ctx context.Context, p svcmod
 		"releaseNotificationConfig": model.ReleaseNotificationConfig(p.ReleaseNotificationConfig), // converted to the struct with json tags (the field is saved as json in the database)
 		"createdAt":                 p.CreatedAt,
 		"updatedAt":                 p.UpdatedAt,
-		"githubRepositoryURL":       p.GithubRepositoryURL.String(),
+		"githubRepositoryURL":       p.GithubRepositoryURL.String(), // TODO remove
 	})
 	if err != nil {
 		return fmt.Errorf("failed to create project: %w", err)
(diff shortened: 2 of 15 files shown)


repository/model/project.go
```diff
@@ -14,7 +16,10 @@ type Project struct {
 	Name                      string                    `db:"name"`
 	SlackChannelID            string                    `db:"slack_channel_id"`
 	ReleaseNotificationConfig ReleaseNotificationConfig `db:"release_notification_config"`
-	GithubRepositoryURL       string                    `db:"github_repository_url"`
+	GithubRepositoryURL       string                    `db:"github_repository_url"` // TODO remove this field
+	GithubOwnerSlug           sql.NullString            `db:"github_owner_slug"`
+	GithubRepoSlug            sql.NullString            `db:"github_repo_slug"`
+	GithubRepoURL             sql.NullString            `db:"github_repo_url"`
```
jan-zabloudil, line 22: The RepoURL will be saved to the database because it is directly returned by the GitHub client. This avoids complicating the implementation with dynamic URL generation.

When the GitHub repo URL needs updating, you can simply call `/projects/{id}/github-repository` to update the connection.

-----

Yes it is make sense to generate git tag url dynamically but not the repo url.

repository/project.go
```diff
@@ -144,7 +144,10 @@ func (r *ProjectRepository) UpdateProject(ctx context.Context, projectID uuid.UU
 		"name":                      p.Name,
 		"slackChannelID":            p.SlackChannelID,
 		"releaseNotificationConfig": model.ReleaseNotificationConfig(p.ReleaseNotificationConfig), // converted to the struct with json tags (the field is saved as json in the database)
-		"githubRepositoryURL":       p.GithubRepositoryURL.String(),
+		"githubRepositoryURL":       p.GithubRepositoryURL.String(),                               // TODO remove
+		"githubOwnerSlug":           p.GetGithubOwnerSlug(),
+		"githubRepoSlug":            p.GetGithubRepoSlug(),
+		"githubRepoURL":             p.GetGithubRepoURL().String(),
```
jan-zabloudil, line 150: Because Project has pointer to GithubRepo 

```
type Project {
....
GithubRepo *GithubRepo
}
```

if i would access GithubRepo fields directly, I would risk panic


```

var p svcmodel.Project

p.GithubRepo.OwnerSlug <-- possible panic (null pointer)
```

Therefore I added Getters

service/model/project.go
```diff
@@ -21,30 +21,54 @@ type Project struct {
 	Name                      string
 	SlackChannelID            string
 	ReleaseNotificationConfig ReleaseNotificationConfig
-	GithubRepositoryURL       url.URL
+	GithubRepositoryURL       url.URL // TODO remove this field
 	GithubRepo                *GithubRepo
```
jan-zabloudil, line 25: I want to keep `GithubRepo` struct in the service model, instead of having these fields  directly in project struct.

```
type Project struct {
...
   URL       url.URL
   OwnerSlug string
   RepoSlug  string
}
```

Why?

- I can easily check if GithubRepo is set `p.GithubRepo != nil`
- I have method for setting repo for project `func (p *Project) SetGithubRepo(repo *GithubRepo) {`
- and mainly github client can return github repo as one struct (instead of returning 3 params):
```
func (c *Client) ReadRepository(ctx context.Context, tkn string, rawRepoURL string) (svcmodel.GithubRepo, error) {
```

transport/handler/routes.go
```diff
@@ -59,7 +59,10 @@ func (h *Handler) setupRoutes() {
 					r.Patch("/", middleware.RequireAuthUser(h.updateMemberRole))
 				})
 			})
-			r.Get("/repository/tags", middleware.RequireAuthUser(h.listGithubRepositoryTags))
+			r.Route("/github-repository", func(r chi.Router) {
+				r.Post("/", middleware.RequireAuthUser(h.setGithubRepoForProject))
+				r.Get("/tags", middleware.RequireAuthUser(h.listGithubRepositoryTags))
```
jan-zabloudil, line 64: Just note to myself, rename `listGithubRepositoryTags` -> `listGithubRepoTags`

repository/project.go
```diff
@@ -139,14 +139,7 @@ func (r *ProjectRepository) UpdateProject(ctx context.Context, projectID uuid.UU
...
 
-	_, err = tx.Exec(ctx, query.UpdateProject, pgx.NamedArgs{
-		"id":                        p.ID,
-		"name":                      p.Name,
-		"slackChannelID":            p.SlackChannelID,
-		"releaseNotificationConfig": model.ReleaseNotificationConfig(p.ReleaseNotificationConfig), // converted to the struct with json tags (the field is saved as json in the database)
-		"githubRepositoryURL":       p.GithubRepositoryURL.String(),
-		"updatedAt":                 p.UpdatedAt,
-	})
+	_, err = tx.Exec(ctx, query.UpdateProject, model.ToUpdateProjectArgs(p))
```
jan-zabloudil, line 142: Preparing arguments for this query felt like too many lines to have directly in Repository function, therefore I added `model.ToUpdateProjectArgs`

service/model/project.go
bolshakoff: Yep, sounds fair.

service/project.go
```diff
@@ -154,6 +147,40 @@ func (s *ProjectService) UpdateProject(ctx context.Context, u model.UpdateProjec
...
+		return fmt.Errorf("getting Github token: %w", err)
+	}
+
+	repo, err := s.githubManager.ReadRepository(ctx, tkn, rawRepoURL)
+	if err != nil {
+		return fmt.Errorf("reading github repository: %w", err)
+	}
+
+	_, err = s.repo.UpdateProject(ctx, projectID, func(p model.Project) (model.Project, error) {
+		p.SetGithubRepo(&repo)
```
bolshakoff, line 174: Elegant.

transport/handler/routes.go
bolshakoff: But then also client methods and other places?..
Because indeed, `GithubRepo` / `Repo` sounds better; `Repository` gets confused with repository pattern a bit.

bolshakoff: (optional, minor) Also, it might be a very, very nice to maybe start creating a dictionary in `README`.
Agreeing on one term and stick to it and define it there. Ubiquitous language - DDD principle.
In our case, `GitTag` for example - might be nice to define that practically, it's always GitHub tag...

bolshakoff (approved):

transport/handler/routes.go
jan-zabloudil: I like it 👍 

https://github.com/jan-zabloudil/release-manager/issues/122 added

PR #121, opened 2024-05-29
feat(github-client): add function to dynamically generate git tag URL

jan-zabloudil:
Git tag URL will be generated dynamically.

### DIFF
--- a/github/client.go
+++ b/github/client.go
@@ -230,6 +230,16 @@ func (c *Client) updateRelease(ctx context.Context, tkn string, repo svcmodel.Gi
 	return nil
 }
 
+func (c *Client) GenerateGitTagURL(repo svcmodel.GithubRepo, tagName string) (url.URL, error) {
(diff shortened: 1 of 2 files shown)


github/client.go
```diff
@@ -33,6 +33,16 @@ func (c *Client) GetGithubRepoFromRawURL(rawURL string) (svcmodel.GithubRepo, er
 	return repo, nil
 }
 
+func (c *Client) GenerateGitTagURL(repo svcmodel.GithubRepo, tagName string) (url.URL, error) {
```
jan-zabloudil, line 36: Accepting svcmodel.GithubRepo (instead of url)

bolshakoff: Btw, here as well, as I once pointed out about having separate type for `userID` (= dumb wrapper around `uuid.UUID`), here a part of me also asks for having a type `githubURL`.
If time allows, would be almost a banger for the thesis / project.

bolshakoff (approved):

PR #127, opened 2024-05-29
feat(release svc): add git tag to release creation

jan-zabloudil:
Creating release logic updated:
- checks if git tag actually exists
- git tag is also added to release repo model and tag is saved to the db 

For now only create release func is updated, read functions (showing git tag and git tag url in response) will be added in separate PR once this one is merged.

### DIFF
--- a/service/errors/errors.go
+++ b/service/errors/errors.go
@@ -7,37 +7,38 @@ import (
 )
 
 var (
-	ErrCodeUnauthorizedUnknownUser           = "ERR_UNAUTHORIZED_ACCESS_UNKNOWN_USER"
-	ErrCodeForbiddenInsufficientUserRole     = "ERR_FORBIDDEN_ACCESS_INSUFFICIENT_USER_ROLE"
-	ErrCodeForbiddenInsufficientProjectRole  = "ERR_FORBIDDEN_ACCESS_INSUFFICIENT_PROJECT_ROLE"
-	ErrCodeForbiddenUserNotProjectMember     = "ERR_FORBIDDEN_USER_NOT_PROJECT_MEMBER"
-	ErrCodeUserNotFound                      = "ERR_USER_NOT_FOUND"
-	ErrCodeProjectNotFound                   = "ERR_PROJECT_NOT_FOUND"
-	ErrCodeEnvironmentNotFound               = "ERR_ENVIRONMENT_NOT_FOUND"
-	ErrCodeProjectUnprocessable              = "ERR_PROJECT_UNPROCESSABLE"
-	ErrCodeEnvironmentUnprocessable          = "ERR_ENVIRONMENT_UNPROCESSABLE"
-	ErrCodeEnvironmentDuplicateName          = "ERR_ENVIRONMENT_DUPLICATE_NAME"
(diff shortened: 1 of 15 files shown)


service/model/release.go
```diff
@@ -37,7 +37,7 @@ type Release struct {
 	GitTagURL    url.URL
 }
 
-func NewRelease(input CreateReleaseInput, projectID, authorUserID uuid.UUID) (Release, error) {
+func NewRelease(input CreateReleaseInput, gitTagURL url.URL, projectID, authorUserID uuid.UUID) (Release, error) {
```
bolshakoff, line 40: Here I'd definitely expect only `CreateReleaseInput`.

I think something like `NewCreateReleaseInput` function (somewhere near `Release` model) would probably the cleanest way. 🤔

service/release.go
```diff
@@ -51,11 +54,32 @@ func (s *ReleaseService) Create(
 		return model.Release{}, fmt.Errorf("getting project: %w", err)
 	}
 
-	rls, err := model.NewRelease(input, projectID, authorUserID)
+	// Fail early by first calling functions that depend only on user input to validate it.
+	// Generate the Git tag URL and create a new release based solely on user input.
+	gitTagURL, err := s.githubManager.GenerateGitTagURL(*p.GithubRepo, input.GitTagName)
```
bolshakoff, line 59: nil pointer dereference?

service/release.go
```diff
@@ -51,11 +54,32 @@ func (s *ReleaseService) Create(
 		return model.Release{}, fmt.Errorf("getting project: %w", err)
 	}
 
-	rls, err := model.NewRelease(input, projectID, authorUserID)
+	// Fail early by first calling functions that depend only on user input to validate it.
```
bolshakoff, line 57: Valid intention, but even without nil dereference, had to spend too much time reading this method.
That's much more costly than a couple microseconds of computing.
So, I'd definitely prioritize readability and smoothness, for example something hierarchical like:
1. Organization: GitHub integration is set + get GitHub token = sort of `githubGuard`
2. Project: GitHub repo is set -> sort of `githubRepoGuard`
3. Release: GitHub tag exists -> sort of `githubTagGuard` 😄 🤷‍♀️

service/release_test.go
```diff
@@ -15,6 +16,7 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
+/*
```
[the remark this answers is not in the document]
bolshakoff: I think `testing.T` should have something like `Skip()` method for such cases. 😉

bolshakoff (approved):

bolshakoff:
Ping me once you get to it; would love to check, if I'm still online.
But I guess it should be fine; so don't want to block merging by not approving. 🤙

service/errors/errors.go
```diff
@@ -182,10 +183,10 @@ func NewGithubIntegrationNotEnabledError() *Error {
 	}
 }
 
-func NewGithubRepoNotConfiguredForProjectError() *Error {
+func NewGithubRepoNotSetForProjectError() *Error {
```
jan-zabloudil, line 186: To keep it consistent, I updated name of the error

bolshakoff (approved):
Yess, reads so much faster.

image

PR #130, opened 2024-05-31
feat: add endpoint for upserting github release

jan-zabloudil:
I will add tests before merging if you approve.

PR #131, opened 2024-05-31
 feat: enable to delete github release when deleting release

jan-zabloudil:
I will add tests before merging if you approve.

### DIFF
--- a/service/release.go
+++ b/service/release.go
@@ -105,9 +105,18 @@ func (s *ReleaseService) Get(ctx context.Context, projectID, releaseID, authorUs
 	return rls, nil
 }
 
-func (s *ReleaseService) Delete(ctx context.Context, projectID, releaseID, authorUserID uuid.UUID) error {
+// Delete deletes a release. If deleteGithubRelease is true, it will also delete associacted GitHub release (if exists).
+// Deleting GitHub release is idempotent, so if the release does not exist on GitHub, it will not return an error.
+func (s *ReleaseService) Delete(ctx context.Context, input model.DeleteReleaseInput, projectID, releaseID, authorUserID uuid.UUID) error {
 	if err := s.authGuard.AuthorizeProjectRoleEditor(ctx, projectID, authorUserID); err != nil {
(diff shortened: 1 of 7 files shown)


service/release_test.go
```diff
@@ -228,10 +228,13 @@ func TestReleaseService_Get(t *testing.T) {
 }
 
 func TestReleaseService_Delete(t *testing.T) {
+	t.Skip("TODO update test case for new implementation")
```
jan-zabloudil, line 231: TODO before merging

transport/handler/release.go
```diff
@@ -45,8 +45,15 @@ func (h *Handler) getRelease(w http.ResponseWriter, r *http.Request) {
...
 func (h *Handler) deleteRelease(w http.ResponseWriter, r *http.Request) {
+	var input model.DeleteReleaseInput
+	if err := util.UnmarshalRequest(r, &input); err != nil {
+		util.WriteResponseError(w, resperrors.NewBadRequestError().Wrap(err).WithMessage(err.Error()))
+		return
+	}
+
 	if err := h.ReleaseSvc.Delete(
 		r.Context(),
+		model.ToSvcDeleteReleaseInput(input),
```
jan-zabloudil, line 56: It might feel like a lot of unnecessary boilerplate code for an input with just one field, but I think it is better to do it this way to be consistent with the rest of the application. Besides, you let GitHub Copilot generate these simple function anyway, so it only takes a minute of work.

service/release.go
```diff
@@ -194,3 +203,30 @@ func (s *ReleaseService) SendReleaseNotification(ctx context.Context, projectID,
 
 	return nil
 }
+
+func (s *ReleaseService) deleteGithubRelease(ctx context.Context, projectID, releaseID, authUserID uuid.UUID) error {
+	rls, err := s.repo.Read(ctx, projectID, releaseID)
```
jan-zabloudil, line 208: I will also change order as you suggested here https://github.com/jan-zabloudil/release-manager/pull/130

bolshakoff (approved):

PR #134, opened 2024-06-02
feat: generate github urls dynamically

jan-zabloudil:
Both github repo url and github tag url are generated dynamically.

 All these changes are included in one PR, but each change is in a separate commit for easier review.


repository/repository.go
```diff
@@ -13,18 +14,23 @@ type querier interface {
 	Query(ctx context.Context, query string, args ...interface{}) (pgx.Rows, error)
 }
 
+type githubURLGenerator interface {
```
bolshakoff, line 17: Nice.

repository/release.go
```diff
@@ -149,7 +152,35 @@ func (r *ReleaseRepository) ListReleasesForProject(ctx context.Context, projectI
...
+	for _, release := range releases {
+		svcRls, err := r.toSvcRelease(release)
+		if err != nil {
+			return nil, err
+		}
+
+		rls = append(rls, svcRls)
+	}
+
+	return rls, nil
```
bolshakoff, line 183: That doesn't feel smooth. And it's a bit confusing (two `toSvcRelease()` methods).
I'd keep the URL generation in the methods where it is needed. Or in the worst case wrap it into a helper here.

And do we need to check the validity?
In case repo slug and owner slug are empty, I'd just generate an empty URL.

bolshakoff (approved):
Not a fan of having two `toSvcRelease()` methods and would try to simplify it.
Otherwise OK. 👌

repository/release.go
jan-zabloudil: I declare empty URL first

```
var tagURL url.URL 
```

Then, if the GitHub owner and repository slugs for the project are set, I generate the real URL for the tag. Otherwise, I return the empty URL.

However, I dislike the idea of generating a URL when the GitHub repository is not set. This can result in URLs like:

```
https://github.com///releases/tag/{tagName}
```

I think there is an improvement for future refactoring. If I have git tag for release, I shouldn't be worry that project repo is not set. User cannot remove repo connection (once project is connected to repository, repository can be only changed). So practically you wont end up in the state where you have git tag but project is not connected to the repository. 

But I think database model should reflect this behaviour as well because now as you can see, I need to handle the case.

----

Regarding `toSvcRelease`/`toSvcReleases`, I will try to move it to helper methods.

bolshakoff: >However, I dislike the idea of generating a URL even when the GitHub repository is not set. This can result in URLs like:

Yes. But my point is, I'd consider moving that logic ("if repo is empty, then return empty URL") to the `GithubClient.GenerateGitTagURLForRelease` method.
But it's really minor now, after the recent change when it's nicely wrapped into an explicit helper. 👌 

(One argument though - if you were to write a test for `GithubClient.GenerateGitTagURLForRelease`, wouldn't you also test these corner cases when the inputs are empty strings? And wouldn't you implement it, i.e. return empty URL, already on that level?)

jan-zabloudil: >(One argument though - if you were to write a test for GithubClient.GenerateGitTagURLForRelease, wouldn't you also test these corner cases when the inputs are empty strings? And wouldn't you implement it, i.e. return empty URL, already on that level?)

Good point 👍. I will merge the current PR and refactor it in a separate PR.

PR #141, opened 2024-06-04
feat(release svc): add deployment information to release notification

jan-zabloudil:


repository/query/scripts/list_deployments_for_project.sql
```diff
@@ -11,6 +11,7 @@ SELECT
     e.id AS env_id,
     e.project_id AS env_project_id,
     e.name AS env_name,
+    e.service_url AS env_service_url,
```
jan-zabloudil, line 14: One fix for the previous PR (service url was not fetched)

service/errors/errors.go
```diff
@@ -295,6 +296,13 @@ func NewDeploymentUnprocessableError() *Error {
 	}
 }
 
+func NewDeploymentNotFoundError() *Error {
```
jan-zabloudil, line 299: Error is not translated in the transport layer because it is never passed to transport layer (for now).

bolshakoff (approved):

PR #143, opened 2024-06-06
feat: delete release when associated git tag is deleted from github repo

jan-zabloudil:


transport/errors/errormapper.go
```diff
@@ -8,24 +8,24 @@ import (
...
-	case isBadRequestError(err):
+	case IsBadRequestError(err):
 		return NewBadRequestError().Wrap(err)
 	default:
 		return NewServerError().Wrap(err)
 	}
 }
 
-func isNotFoundError(err error) bool {
+func IsNotFoundError(err error) bool {
```
jan-zabloudil, line 28: I made those functions in /transport/errors/errrormapper public because there may be cases when I want to check the error type in the handler (as you can see in `webhook` handler)

transport/handler/routes.go
```diff
@@ -94,7 +94,7 @@ func (h *Handler) setupRoutes() {
 		r.Patch("/", middleware.RequireAuthUser(h.updateSettings))
 	})
 
-	h.Mux.Post("/webhooks/github/releases", h.handleGithubReleaseWebhook)
+	h.Mux.Post("/webhooks/github/tags", h.handleGithubReleaseWebhook)
```
jan-zabloudil, line 97: Listening for git tag events, therefore the route has been updated.

repository/release.go
```diff
@@ -128,14 +128,26 @@ func (r *ReleaseRepository) UpdateRelease(
...
 	// But it is added for security reasons
 	// To make sure that the release belongs to the project that is passed from the service
-	result, err := r.dbpool.Exec(ctx, query.DeleteRelease, pgx.NamedArgs{
+	return r.deleteRelease(ctx, query.DeleteRelease, pgx.NamedArgs{
 		"projectID": projectID,
 		"releaseID": releaseID,
 	})
+}
+
+func (r *ReleaseRepository) deleteRelease(ctx context.Context, deleteQuery string, args pgx.NamedArgs) error {
```
bolshakoff, line 149: Absolutely understand the strive, but be careful with this.
Yes, you refactored it, with its benefits, etc., but simple metrics to be aware of:
- I spent like 2 minutes here (instead of 10 seconds in case it was repeated code).

Next time it might be less, but still, more than 10 seconds.

And then if you have a team of 10 people, over a year, it's like hours of difference in terms of time spent reading and reasoning about the code.

You can leave it like that, but might be an interesting thing to point out in the thesis.

Also, we are in Go. And Go was designed to scale, mainly in terms of number of engineers using it (you have a simple, unambiguous style guides etc.).

And again, we are in Go, and one idiomatic way to solve such things (from my understanding and experience, plus from what I saw in huge programs written in Go, for example [GCP Terraform provider by Google itself](https://github.com/hashicorp/terraform-provider-google?tab=readme-ov-file#developing-the-provider))..
One idiomatic way to solve this, is to generate the code.

But that's a whole different story of course.

And here it's obviously an overkill. 😄

bolshakoff (approved):

repository/release.go
jan-zabloudil: Thanks!

PR #144, opened 2024-06-07
refactor(github client): validate parameters when generating URL

jan-zabloudil:


github/client.go
```diff
@@ -226,6 +226,10 @@ func (c *Client) updateRelease(ctx context.Context, tkn string, repo svcmodel.Gi
 }
 
 func (c *Client) GenerateRepoURL(ownerSlug, repoSlug string) (url.URL, error) {
+	if ownerSlug == "" || repoSlug == "" {
+		return url.URL{}, errors.New("empty owner or repo slug")
+	}
```
jan-zabloudil, line 231: If an empty parameter is provided, the GitHub client returns an error. 

Initially, I considered returning an empty URL without an error. However, it makes more sense to return an error because you should either not attempt to generate a URL at all in the service/repository, or when you call a function for URL generation, you should ensure that you receive either an error or a non-empty URL.

No service error is returned because this error is not expected. When generating a URL for my GitHub repository or git tags, all parameters should be present when calling the function. This is similar to receiving an error when parsing a URL.

repository/project.go
```diff
@@ -86,7 +86,12 @@ func (r *ProjectRepository) readProject(ctx context.Context, q querier, query st
 		return svcmodel.Project{}, err
 	}
 
-	return r.toSvcProject(p)
+	repo, err := r.toSvcGithubRepo(p)
+	if err != nil {
+		return svcmodel.Project{}, fmt.Errorf("failed to convert github repo: %w", err)
+	}
+
+	return model.ToSvcProject(p, repo), nil
```
jan-zabloudil, line 94: When fetching project/projects from repository. I also had two functions:

```
r.toSvcProject
r.toSvcProjects
```

These functions were removed (to follow your suggestion for release repo). There is only one helper method ` r.toSvcGithubRepo`.

repository/release.go
```diff
@@ -76,7 +75,7 @@ func (r *ReleaseRepository) readRelease(ctx context.Context, q querier, readQuer
 		return svcmodel.Release{}, err
 	}
 
-	tagURL, err := r.generateGitTagURLForRelease(rls)
+	tagURL, err := r.githubURLGenerator.GenerateGitTagURL(rls.GithubOwnerSlug.String, rls.GithubRepoSlug.String, rls.GitTagName)
```
jan-zabloudil, line 78: When fetching a release, I should always have the GitHub owner and slug. Therefore, I call GenerateGitTagURL directly without checking the values first. If empty values are provided, the GitHub client returns an error.

bolshakoff (approved):
image

PR #167, opened 2024-09-07
feat: add release attachments

jan-zabloudil:
**Objective:**

As an engineer, I want to be able to attach files to a release so that I can include relevant documents, such as OpenAPI documentation. Notes:
- Each release can have multiple files attached.
- User creates release first. Once release is created, files can be attached.

**Implementation:**

[Supabase Storage](https://supabase.com/storage) will be used for storing and serving files.

**Expected flow:**

- Client requests a pre-signed URL for uploading a file.
- File is uploaded via the pre-signed URL, returning a storage object key.
- A POST request is made to create a new attachment (request body includes object key).

**Planned PRs:**

To keep pull requests manageable, several PRs will be created:
- Update release models in all layers to include attachments (this PR).
- Create an endpoint for generating pre-signed URLs for uploads.
- Implement POST `/projects/{projectID}/release/{releaseID}/attachments`.
- Implement DELETE `/projects/{projectID}/release/{releaseID}/attachments/{attachmentID}`.
- Implement a mechanism to delete files when the related entity (release or the entire project) is deleted.
- Implement caching for signed URLs for downloads.

I try to introduce as simple implementation as possible (keeping KISS in my mind). First time implementing file uploads via signed urls 🤞 happy to hear your feedback.


service/model/release.go
```diff
@@ -50,6 +50,15 @@ type Release struct {
...
+	Attachments  []ReleaseAttachment
+}
+
+type ReleaseAttachment struct {
+	ID        uuid.UUID
+	Name      string
+	FilePath  string
+	URL       url.URL
+	CreatedAt time.Time
 }
```
jan-zabloudil, line 62: - To keep things simple, there is simple `ReleaseAttachment` struct. 
- If the app needs to store more types of files in the future, it would be a good opportunity to refactor and introduce a general `File model` (IMHO).
- Although `FilePath` could serve as a natural identifier, I thing it is more flexible and practical to use UUID as ID.

Note to FilePath field:
_Since the service model doesn't concern itself with where the files are stored, I chose the name FilePath for the field. While cloud storage services often use terms like 'storage object key' or 'storage object ID,' I believe FilePath is more general and does not imply that a specific type of storage is being used_

repository/model/release.go
```diff
@@ -20,18 +20,41 @@ type Release struct {
...
+) (svcmodel.Release, error) {
+	tagURL, err := tagURLGenerator(rls.GithubOwnerSlug.String, rls.GithubRepoSlug.String, rls.GitTagName)
 	if err != nil {
-		return svcmodel.Release{}, fmt.Errorf("failed to generate tag URL: %w", err)
+		return svcmodel.Release{}, fmt.Errorf("generating a git tag URL: %w", err)
+	}
+
+	attachments := make([]svcmodel.ReleaseAttachment, 0, len(rls.Attachments))
+	for _, a := range rls.Attachments {
+		u, err := fileURLGenerator(a.FilePath)
```
jan-zabloudil, line 52: The repository defines the following interface:

```
type fileURLGenerator interface {
    GenerateFileURL(fileObjectKey string) (url.URL, error)
}
```

- A storage client that satisfies this interface is injected into the repository. 
- A repo model function then accepts this URL generator function as a parameter, enabling the generation of file URLs when converting the repository model to the service model.

This approach follows the same principle previously used for generating GitHub URLs.

supabase/migrations/20240906043335_release_attachments.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE public.release_attachments (
+    release_id UUID NOT NULL REFERENCES public.releases ON DELETE CASCADE,
+    attachment_id UUID NOT NULL,
+    name TEXT NOT NULL,
+    file_path TEXT NOT NULL,
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL,
+    PRIMARY KEY (release_id, attachment_id),
+    CONSTRAINT unique_file UNIQUE (file_path)
+)
```
jan-zabloudil, line 9: If needed in the future, the database model can be migrated to include a `file` entity. 

The `release_attachments` table would then have `release_id` and `file_id` fields (along with any other attachment-specific fields).

storage/client.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+func (c *Client) GenerateFileURL(fileKey string) (url.URL, error) {
+	bucketID, filePath := util.ExplodeFileKey(fileKey)
+	if bucketID == "" || filePath == "" {
+		return url.URL{}, fmt.Errorf("parsing file key %s: %w", fileKey, errInvalidFileKey)
+	}
+
+	resp, err := c.client.Storage.From(bucketID).CreateSignedURLForDownload(filePath, signedURLExpiration)
```
jan-zabloudil, line 37: Calling the Supabase Storage API and generating a new signed URL each time a release  is read from the database could result in significant traffic to the Storage API. 

I thing It would make sense to generate signed URLs with a longer expiration time (several hours, at least?) and cache them (using Redis for example).

go.mod
```diff
@@ -44,4 +44,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-replace github.com/nedpals/supabase-go => github.com/jan-zabloudil/supabase-go v0.5.0
+replace github.com/nedpals/supabase-go => github.com/jan-zabloudil/supabase-go v0.6.0
```
jan-zabloudil, line 47: A new tag has been published with fixes for generating signed URLs

https://github.com/jan-zabloudil/supabase-go/releases/tag/v0.6.0

jan-zabloudil:
I need to store file paths without including the bucket ID and instead pass the bucket ID as an environment variable. I will be opening a new PR soon.

PR #168, opened 2024-09-09
feat: add release attachments

jan-zabloudil:
**Objective:**

As an engineer, I want to be able to attach files to a release so that I can include relevant documents, such as OpenAPI documentation. 

Notes:
- Each release can have multiple files attached.
- User creates release first. Once release is created, files can be attached.

**Implementation:**

[Supabase Storage](https://supabase.com/storage) will be used for storing and serving files.

**Expected flow:**

- Client requests a pre-signed URL for uploading a file.
- File is uploaded via the pre-signed URL, returning a storage object key.
- A POST request is made to create a new attachment (request body includes object key).

**Planned PRs:**

To keep pull requests manageable, several PRs will be created:
- Update release models in all layers to include attachments (this PR).
- Create an endpoint for generating pre-signed URLs for uploads.
- Implement POST `/projects/{projectID}/release/{releaseID}/attachments`.
- Implement DELETE `/projects/{projectID}/release/{releaseID}/attachments/{attachmentID}`.
- Implement a mechanism to delete files when the related entity (release or the entire project) is deleted.
- Implement caching for signed URLs for downloads.

I try to introduce as simple implementation as possible (keeping KISS in my mind). First time implementing file uploads via signed urls 🤞 happy to hear your feedback.


service/model/release.go
```diff
@@ -50,6 +50,15 @@ type Release struct {
 	UpdatedAt    time.Time
 	GitTagName   string
 	GitTagURL    url.URL
+	Attachments  []ReleaseAttachment
+}
+
+type ReleaseAttachment struct {
```
jan-zabloudil, line 56: - To keep things simple, there is only `ReleaseAttachment` struct.
- If the app needs to store more types of files in the future, it would be a good opportunity to refactor and introduce a general File model as well (IMHO).
- Although `FilePath` could serve as a natural identifier, I thing it is more flexible and practical to use UUID as ID.

repository/model/release.go
```diff
@@ -20,18 +20,41 @@ type Release struct {
...
+) (svcmodel.Release, error) {
+	tagURL, err := tagURLGenerator(rls.GithubOwnerSlug.String, rls.GithubRepoSlug.String, rls.GitTagName)
 	if err != nil {
-		return svcmodel.Release{}, fmt.Errorf("failed to generate tag URL: %w", err)
+		return svcmodel.Release{}, fmt.Errorf("generating a git tag URL: %w", err)
+	}
+
+	attachments := make([]svcmodel.ReleaseAttachment, 0, len(rls.Attachments))
+	for _, a := range rls.Attachments {
+		u, err := fileURLGenerator(a.FilePath)
```
jan-zabloudil, line 52: The repository defines the following interface:

```
type fileURLGenerator interface {
    GenerateFileURL(filePath string) (url.URL, error)
}
```

- A storage client that satisfies this interface is injected into the repository. 
- A repo model function then accepts this URL generator function as a parameter, enabling the generation of file URLs when converting the repository model to the service model.

This approach follows the same principle previously used for generating GitHub URLs.

supabase/migrations/20240906043335_release_attachments.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE public.release_attachments (
```
jan-zabloudil, line 1: If needed in the future, the database model can be migrated to include a `file` entity. 

The `release_attachments` table would then have `release_id` and `file_id` fields (along with any other attachment-specific fields).

storage/client.go
```diff
@@ -0,0 +1,39 @@
...
+
+func NewClient(client *supabase.Client, bucket string) *Client {
+	return &Client{
+		client: client,
+		bucket: bucket,
+	}
+}
+
+func (c *Client) GenerateFileURL(filePath string) (url.URL, error) {
+	resp, err := c.client.Storage.From(c.bucket).CreateSignedURLForDownload(filePath, signedURLExpiration)
```
jan-zabloudil, line 28: Calling the Supabase Storage API and generating a new signed URL each time a release  is read from the database could result in heavy traffic to the Storage API (and increased response time). 

I think It would make sense to generate signed URLs with a longer expiration time (several hours, at least?) and cache them (using some in-memory package, e.g. [freecache](https://github.com/coocood/freecache), I think using Redis or something similar would be overkill, It's acceptable if the cache is lost when the application stops).

Another optimization could be generating signed URLs for all the files needed for releases in a single request. Supabase Storage offers an endpoint that supports this approach.

On the other hand, we're still dealing with a relatively small number of files in the case of ReleaseManager, so I wouldn't complicate it too much. Love to know your opinion.

go.mod
```diff
@@ -44,4 +44,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-replace github.com/nedpals/supabase-go => github.com/jan-zabloudil/supabase-go v0.5.0
+replace github.com/nedpals/supabase-go => github.com/jan-zabloudil/supabase-go v0.6.0
```
jan-zabloudil, line 47: A new tag has been published with fixes for generating signed URLs

https://github.com/jan-zabloudil/supabase-go/releases/tag/v0.6.0

TomasKocman (approved):

Issue #177, opened 2024-09-16
REST API - reduce nested resources


jan-zabloudil:
**Problem**

Currently, all endpoints related to releases include the project in the path. For example:
`/projects/{project_id}/releases/{release_id}`

Implementing features like release attachments has become more complex with paths like:
`DELETE /projects/{project_id}/releases/{release_id}/attachments/{attachment_id}`

**Solution**

To simplify this, the project identifier will be removed from the path when working with a single release:

`/projects/{project_id}/releases/{release_id}` → `/releases/{release_id}`
`/projects/{project_id}/releases/{release_id}/slack-notifications` → `/releases/{release_id}/slack-notifications`
`/projects/{project_id}/releases/{release_id}/attachments/{attachment_id}` → `/releases/{release_id}/attachments/{attachment_id}`

**Benefits**

This change simplifies the design of the endpoints and also the implementation in the service and repository layers (no need to check if given project exists and is associated with given release etc).

PR #186, opened 2024-09-25
feat(authorization svc): add methods for authorizing release actions

jan-zabloudil:
This PR is related to #177.

As proposed in the issue, the project ID will no longer be part of the release endpoint routes. Therefore the authorization service must provide methods to authorize release actions based solely on the release ID and user ID.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -54,7 +54,8 @@ type settingsRepository interface {
 
 type releaseRepository interface {
 	CreateRelease(ctx context.Context, r model.Release) error
-	ReadRelease(ctx context.Context, projectID, releaseID uuid.UUID) (model.Release, error)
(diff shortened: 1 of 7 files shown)


service/service.go
```diff
@@ -54,7 +54,8 @@ type settingsRepository interface {
 
 type releaseRepository interface {
 	CreateRelease(ctx context.Context, r model.Release) error
-	ReadRelease(ctx context.Context, projectID, releaseID uuid.UUID) (model.Release, error)
+	ReadRelease(ctx context.Context, releaseID uuid.UUID) (model.Release, error)
+	ReadReleaseForProject(ctx context.Context, projectID, releaseID uuid.UUID) (model.Release, error)
```
jan-zabloudil, line 58: This is only temporary. Once the refactoring in #177 is complete, all release repository functions that handle a single release will only require the release ID.

